PR #16, opened 2026-09-02
feat(fmt): replace the whitespace pass with a native Flow printer

ubugeeei:
`uf fmt` was 87 lines that trimmed trailing whitespace and expanded leading
tabs. Replace it with a real formatter built on two linear passes over a flat
token vector: no syntax tree, no per-node allocation, no recursion.

- `lexer.rs`: a lossless single-pass tokenizer. Every byte belongs to exactly
  one token, so concatenating the token slices reproduces the input. Handles
  line/block/JSDoc comments, string escapes and line continuations, nested
  `${}` templates, regex-vs-division via the previous-significant-token rule
  (including the `}` and `)` cases that need brace and paren classification),
  JSX text/tags/containers, every numeric literal form, Unicode identifiers
  and Flow type punctuation.
- `printer.rs`: an analysis pass that records spacing, indentation and group
  extents, then an emit pass that writes the output. Indentation follows the
  groups that actually break at their own level, which is what makes hugged
  callbacks and JSX children land where Prettier puts them. Honours every
  `FmtConfig` field, explodes groups that overflow `lineWidth`, and preserves
  comments verbatim.
- `uf_config`: new `fmt.maxBlankLines` (default 1).

Three invariants are enforced by tests: formatting is idempotent, it never
changes the token stream (only trivia, quote style and statement-terminating
semicolons), and no input panics or hangs — truncated literals, unpaired
delimiters, a 1 MB line and 10,000-deep nesting all format. Both passes use
explicit stacks: unbounded recursion over attacker-controlled nesting is a
well-worn crash vector for formatters.

Adds 12 golden fixtures driven by `tests/fixtures.rs` and a criterion bench.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/16"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790918617&installation_model_id=422833&pr_number=16&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F16&signature=334ea48bf04ab873da02ba2d5e5a5691e71541ee604f9785e55d776b1c91a9cf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive formatting for Flow JavaScript and JSX, including comments, templates, regular expressions, types, and nested structures.
  * Added configurable maximum blank lines, line width, indentation, quote style, and semicolon handling.
  * Added plugin configuration support and development settings for filesystem access, hosts, and origins.
  * Added automatic line wrapping and consistent indentation while preserving source content and comments.
  * Added clear validation when indentation exceeds the supported limit.

* **Bug Fixes**
  * Improved handling of line endings, byte-order marks, and deeply nested code without stack overflow.

* **Tests**
  * Added extensive formatting fixtures, fidelity checks, idempotence tests, and performance benchmarks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1546,6 +1546,8 @@ dependencies = [
 name = "uf_fmt"
 version = "0.1.0"
 dependencies = [
+ "criterion",
+ "memchr",
  "similar-asserts",
  "thiserror",
  "uf_config",
--- a/crates/uf_config/src/lib.rs
+++ b/crates/uf_config/src/lib.rs
@@ -788,6 +788,7 @@ impl Default for EnvConfig {
 pub struct FmtConfig {
     pub indent_width: u8,
     pub line_width: u16,
+    pub max_blank_lines: u8,
     pub quotes: QuoteStyle,
     pub semicolons: bool,
 }
@@ -797,6 +798,7 @@ impl Default for FmtConfig {
         Self {
             indent_width: 2,
             line_width: 100,
+            max_blank_lines: 1,
             quotes: QuoteStyle::Double,
             semicolons: true,
         }
--- a/crates/uf_config/src/tests.rs
+++ b/crates/uf_config/src/tests.rs
@@ -432,3 +432,19 @@ export default defineConfig({
 
     assert!(json5::from_str::<UniflowedConfig>(&object).is_err());
 }
+
+#[test]
+fn fmt_config_reads_max_blank_lines_from_the_config_file() {
+    let source = r#"
+        export default {
+          fmt: { maxBlankLines: 0, indentWidth: 4 },
+        };
+    "#;
+
+    let object = extract_config_object(source).expect("object");
+    let parsed: UniflowedConfig = json5::from_str(&object).expect("config");
+
+    assert_eq!(parsed.fmt.max_blank_lines, 0);
+    assert_eq!(parsed.fmt.indent_width, 4);
+    assert_eq!(parsed.fmt.line_width, 100);
+}
--- a/crates/uf_fmt/Cargo.toml
+++ b/crates/uf_fmt/Cargo.toml
@@ -10,7 +10,13 @@ authors.workspace = true
 
 [dependencies]
 uf_config = { path = "../uf_config" }
+memchr.workspace = true
 thiserror.workspace = true
 
 [dev-dependencies]
+criterion.workspace = true
 similar-asserts.workspace = true
+
+[[bench]]
+name = "fmt"
+harness = false
--- a/crates/uf_fmt/benches/fmt.rs
+++ b/crates/uf_fmt/benches/fmt.rs
@@ -1,0 +1,90 @@
+use std::hint::black_box;
+
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use uf_config::FmtConfig;
+use uf_fmt::{format_source, lexer::tokenize};
+
+/// Build a synthetic Flow + JSX source file of roughly `components` components.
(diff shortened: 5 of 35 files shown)


ubugeeei:
Rebased onto `main` and resolved the conflict with #22.

`#[non_exhaustive]` on the config sections means `FmtConfig { … }` literals no
longer compile outside `uf_config`, which is what #22 traded for letting these
structs gain fields without a major bump — and `max_blank_lines` is exactly such
a field. The 18 literals in this PR's tests and fixtures now go through a
`config_with(|config| …)` helper that starts from `Default::default()`, so the
next knob added to `FmtConfig` will not touch this file at all.

Re-verified after the rebase, all on the pinned 1.98.0 toolchain:

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` — **851 passed, 0 failed**
- `cargo bench --workspace --no-run`

ubugeeei:
Both CodeRabbit findings were real. Fixed rather than resolved.

**1. A body brace after a tuple return type was classified as an object literal.**

`hook useSelection(): [string, (next: string) => void] { … }` came out with a
`;` after the closing brace, because the token before the brace is `]` and
`needs_semicolon` then emitted a terminator for what it thought was an object.
The same reasoning that already special-cases `>` for
`function f(): Promise<void> {` applies: `[1, 2] {}` is not an expression, so
nothing valid puts an object literal directly after `]`.

**Two golden fixtures had the extra semicolon baked in as expected output**,
which is how it survived — a fixture that records a bug turns the bug into the
specification. Both are regenerated and now carry exactly as many semicolons as
their input (react_component 10/10, flow_types 15/15).

The token-preservation invariant did not catch it either, because it
deliberately permits statement-terminating semicolons to change. The new tests
compare semicolon counts against the input directly, which is the property that
actually holds, and **they fail with the fix reverted** — verified.

**2. The malformed-input test only checked idempotence.**

You are right that a stable token-changing rewrite passes that. Full token
preservation cannot hold on malformed input — the printer emits one final
newline, and for a source ending inside an unterminated comment that newline
lands *inside* the comment token — so the corpus now asserts the guarantee that
does hold: no token dropped or invented. Semicolons are filtered for the same
reason the strong check filters them; `</div>` alone is a statement, and adding
its terminator is documented behaviour.

The `github-code-quality` threads on the fixtures are unused variables and
use-before-declaration inside `awkward_comments` and `template_literals`. Those
files are deliberately awkward — they exist to feed the formatter code that a
tidy file would never contain — and they are never executed.

Re-verified: `cargo test --workspace` **1434 passed, 0 failed**; fmt and
`clippy --workspace --all-targets -D warnings` clean.

PR #54, opened 2026-09-02
format with Flow's own layout generator instead of a token heuristic

ubugeeei:
**Draft — one decision needed before this can land. See the bottom.**

`uf_fmt` works from its own lexer and has to *guess* at the grammar. That guessing is what produced the defects:

- `<T extends U>` printed as `< T extends U >`, because a speculative scan could not tell a type-parameter bracket from a comparison
- a tuple return type grew a stray `;`, because a brace classifier read a type as a block

Both were found in generated code, and the router codegen has the first one **baked into its source literal** so the generated file survives `uf fmt --check`.

The Flow Rust port already ships `js_layout_generator` and `pretty_printer` — the code `flow format` runs. Parse with `flow_parser`, lay out, print. An AST printer cannot make either mistake, because it is never guessing.

Verified against the syntax the old formatter mangled:

| input | Flow's output |
|---|---|
| `declare export function route<Path extends RoutePath>(path: Path): string;` | unchanged ✅ |
| `export type Cov<out P> = (P) => void;` | unchanged ✅ |
| `export type A = {\| readonly n: string \|};` | unchanged ✅ |
| `component Button(label: string) renders React.Node { ... }` | reformatted correctly ✅ |
| `const   m:Map<string,number>   =   new Map( );` | `const m: Map<string, number> = new Map();` ✅ |
| `const ok=a<b&&c>d;` | `const ok = a < b && c > d;` ✅ |

## Two things this had to solve

**Stack exhaustion.** Flow's parser and layout generator both recurse on the call stack. I measured it: **500 levels of nesting succeed, 1000 aborts the process.** A formatter that aborts is a denial of service on whatever it is asked to format — and `docs/security.md` has a row for exactly this class. Formatting now runs on a thread with a fixed 16 MB stack, so the accepted depth is not a property of who called, and nesting deeper than 256 is refused with a typed error.

**The `@flow` pragma.** `preserve_docblock: false` drops it outright — the type checker reads it, so that is not an option. `true` keeps it but leaves a blank line at the head of a file that has no docblock. The pragma stays; the leading blank line is trimmed.

## Status: 93 of 122 tests pass

The remaining 29 assert the *old* formatter's conventions. Flow's differ:

| old uf | Flow |
|---|---|
| `type U = \| A \| B & C;` | `type U = A \| (B & C);` — explicit precedence |
| `const xs = [\n  1,\n  2,\n];` | `const xs = [1, 2];` — short arrays inline |
| `(\n  <ul>…</ul>\n)` | `<ul>…</ul>` — no parens around multi-line JSX |
| `// @flow` then code | `// @flow`, blank line, then code |

I judge Flow's output better in each case — the precedence parens especially. But this is a decision about uf's house style, not a bug, so I have not made it.

## The decision

Adopting Flow's formatter means:

1. **Reformatting every source file in the repository** to Flow's conventions.
2. **`FmtConfig` loses four options.** Flow's printer is fixed at 80 columns and two-space indent, with no setting for line width, indent width, blank-line runs, or semicolons. `uf.config.js` advertises all four. This PR makes a non-default value a loud typed error rather than silently printing something else, and changes the `lineWidth` default from 100 to 80 — but the honest end state is removing the fields.

Both follow directly from "use the official assets", and both are user-facing, so they are yours to call. Say the word and I will finish the expectations and the config.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/54"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790938459&installation_model_id=422833&pr_number=54&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F54&signature=3e23744d267e419df0c6a008bd586af542b2a2fa3cf18a27f95551b00c08a46a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Closing: the approach is wrong for the requirement.

This built on Flow's `js_layout_generator`, which is Flow's house style, **not Prettier's**. Measured difference, for one:

| Flow | Prettier |
|---|---|
| `const a = <ul>…</ul>;` | `const a = (\n  <ul>…</ul>\n);` |

Since the requirement is 100% Prettier compatibility, a printer that agrees with Flow instead of Prettier is not a partial answer, it is the wrong one.

What carries over and will be reused:
- **Parse with the official Flow Rust port.** Grammar is correct, verified on `<T extends U>`, `<out P>`, `{| readonly n |}`, `component`/`hook`.
- **Stack bound.** Flow's parser recurses; measured, 500 levels of nesting succeed and 1000 aborts the process. Formatting must run on a fixed stack with a depth cap, or `uf fmt` is a denial of service on whatever it is asked to format.
- **`preserve_docblock: false` drops the `// @flow` pragma** the type checker reads.

Replacement: Prettier's Doc IR (`group`/`indent`/`line`/`softline`) and its `fits` algorithm in Rust, over the Flow AST, verified by differential testing against `prettier --parser=flow` rather than hand-written expectations — hand-written expectations cannot demonstrate 100% compatibility.

PR #111, opened 2026-09-04
feat(relay): Relay, re-exported by name — and the environment nobody should rewrite (#110)

ubugeeei:
`@uniflowed/relay` was five functions that threw. It is the real `react-relay`
now, in exactly the shape `@uniflowed/react` has for `react`: every export
named individually, so the surface is explicit and a bundler can drop what an
application does not touch, and Relay's own Flow types come through unchanged.

That is the red lines' answer rather than a shortcut. Relay is a store, a
normaliser, a compiler and a decade of work on cache consistency; a uf-shaped
reimplementation would be a worse Relay that uf then had to keep up with. uf
owns orchestration, not implementation — so the library is Meta's, and what uf
adds is the toolchain around it.

`@uniflowed/graphql` is the twenty lines every Relay application writes before
it can use Relay: a network layer that posts an operation, a store, and an
environment holding them together. It exists because that boilerplate is not a
design decision — every project writes the same thing, gets the response shape
slightly wrong, and finds out when a GraphQL error is rendered as `null`. A
response carrying `errors` with a 200 throws `GraphQlResponseError` here, which
is the whole reason to have written it once.

It hands back Relay's own `Environment`, so it is not a wrapper: everything
Relay can do a uf project can do, and a project that outgrows the helper
replaces the call rather than escaping a framework. That is red line 8, and
this is what it looks like when it holds.

The fetch client is explicit rather than a global, because uf does not override
`globalThis.fetch` and a server rendering a request often has to forward that
request's credentials — which a global cannot know about. Each call builds a
fresh store, because one environment per request is the only way two users on a
server do not see each other's data.

Twelve tests, run against the real Relay runtime rather than a stand-in: the
request that goes out, the operation name that makes a server's logs readable,
and the identity of every re-exported binding — `uf.useFragment` has to *be*
Relay's function, or an application is holding two Relays.

Co-authored-by: Claude Opus 5 <noreply@anthropic.com>

---

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/111"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791099136&installation_model_id=422833&pr_number=111&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F111&signature=c95f59253cd0ea4404511e14cc7d29eb686b760c3bf69e41a610da5b1df0f798"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added configurable formatting for JSON, CSS, and TypeScript files using Biome, Prettier, or no formatter.
  - Added support for project-local formatter installations and shared formatting settings.
  - `uf explain fmt` now displays the selected non-Flow formatter and command details.

- **Bug Fixes**
  - `uf fmt` now clearly distinguishes formatter failures from unsupported files.
  - Linting excludes JSON, CSS, and TypeScript files handled by the external formatter.

- **Documentation**
  - Updated architecture and project guidance to reflect configurable formatter providers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3398,10 +3398,13 @@ dependencies = [
 name = "uf_fmt"
 version = "0.0.0-alpha.2"
 dependencies = [
+ "camino",
+ "compact_str",
  "criterion",
  "memchr",
  "serde_json",
  "similar-asserts",
+ "tempfile",
  "thiserror",
  "uf_config",
  "uf_flow",
--- a/brand/tokens.css
+++ b/brand/tokens.css
@@ -1,15 +1,15 @@
 :root {
-  --uf-color-cyan-500: #35D6F6;
-  --uf-color-blue-500: #2677FF;
-  --uf-color-indigo-500: #5C49FF;
-  --uf-color-violet-500: #8F4BFF;
-  --uf-color-magenta-500: #D84BFF;
-  --uf-color-ink-900: #0F172A;
+  --uf-color-cyan-500: #35d6f6;
+  --uf-color-blue-500: #2677ff;
+  --uf-color-indigo-500: #5c49ff;
+  --uf-color-violet-500: #8f4bff;
+  --uf-color-magenta-500: #d84bff;
+  --uf-color-ink-900: #0f172a;
   --uf-color-slate-600: #475569;
-  --uf-color-mist-50: #F8FAFC;
-  --uf-color-success-500: #34D399;
-  --uf-color-warning-500: #F59E0B;
-  --uf-color-danger-500: #EF4444;
+  --uf-color-mist-50: #f8fafc;
+  --uf-color-success-500: #34d399;
+  --uf-color-warning-500: #f59e0b;
+  --uf-color-danger-500: #ef4444;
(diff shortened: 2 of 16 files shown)


ubugeeei:
Superseded by #112, which contains this commit plus the lint fixes, the star re-exports and the loader extraction.

PR #115, opened 2026-09-04
feat(test): a clock a test controls, and the namespace `uft`

ubugeeei:
### refactor(test): the namespace is `uf`, not `vi`

Borrowing Vitest's name for it was claiming something uf has not earned. The
*shape* is what a migration needs — `fn`, `spyOn`, `stubEnv`, `useFakeTimers`
doing what they do elsewhere — and that is unchanged; only the object they hang
off is uf's own.

Still a namespace rather than loose named exports, and for a concrete reason:
several of these names are generic enough to collide in one test file.
`@uniflowed/testing` re-exports both this package and `@uniflowed/react-testing`,
and both have a `waitFor`.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### feat(test): a clock a test controls

A test about "after five minutes the session expires" should not take five
minutes, and a test that waits a real 50ms to watch a debounce is a test that
fails on a loaded machine. `vi.useFakeTimers` replaces the scheduling globals
with a queue the test advances by hand, so elapsed time is an input rather than
something to wait for.

`setTimeout`, `setInterval`, `setImmediate`, their `clear` partners, `Date.now`
and `new Date()`. Not `queueMicrotask` and not promises: a microtask is not
scheduled *in time* — it runs at the end of the current turn — and faking it
would let a test claim a promise resolved "after 100ms" when the two have
nothing to do with each other. `advanceTimersByTimeAsync` is the honest way to
let a callback's `await` continue before the next timer fires, and it has the
test that shows the difference.

Details that decide whether a fake clock is trustworthy, each with a test:
timers due at the same instant fire in the order they were scheduled; an
advance lands on the instant it was asked for even when nothing was due there,
so two advances of 50 are one of 100; a timer scheduled *by* a timer inside the
same advance still fires; a zero-delay interval advances by a tick, because one
scheduled at the same instant forever would never let the clock move; and
`setSystemTime` moves the wall clock without firing anything.

`runAllTimers` on an interval would never terminate, so it throws after ten
thousand — an unbounded hang has to be killed, and a failure names the problem.
`runOnlyPendingTimers` is the way to drain one.

The fake `Date` is a subclass rather than a wrapper, so `instanceof Date` still
holds and every method comes along unchanged, and only the no-argument
constructor reads the clock — `new Date("2020-05-05")` is constructing a
specific date and has nothing to do with "now".

Twenty three tests, each installing and removing the clock itself: a leaked
fake clock is the failure that matters most, because the next file's
`setTimeout` never fires and the run hangs with no explanation.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

## Summary by CodeRabbit

- **New Features**
  - Added fake-timer controls, including clock advancement, pending-timer execution, timer counting, and mocked system time.
  - Exposed `RunawayTimersError` for detecting non-terminating timer execution.
  - Renamed the testing namespace from `vi` to `uft` across the public testing API.

- **Bug Fixes**
  - Improved React compiler detection so variables named `hook` or `component` are not mistaken for declarations.
  - Preserved correct handling of genuine hook and component declarations.


ubugeeei:
Both findings were real, and both are fixed in `fix(test): the faked Date is a proxy, not a subclass`.

A `class FakeDate extends Real` got three things wrong, all of them only while fake timers were installed:

- `Date()` without `new` returns a string in every runtime; a class cannot be called that way at all, so it threw `TypeError`.
- `before instanceof Date` was false for a date built before the clock was faked — the object's chain runs through the real `Date`, and a subclass's prototype is not in it. `setSystemTime` branched on exactly that, took the number branch, and set the clock to an object, after which no timer ever came due.
- `Date.name` was `"FakeDate"` (found by the regression tests, not in the review).

A `Proxy` over the real constructor fixes all three at once: no prototype of its own, so `instanceof` is the real check; an `apply` trap restores the callable form; everything but `now` reaches through to the real constructor. `setSystemTime` no longer asks `instanceof` at all — anything that can say what instant it is, is an instant.

Four regression tests added, written red first. 432 passing.

Issue #151, opened 2026-09-05
fmt: an `interface { … }` type separates its members with `;` where the hermes plugin uses `,`


ubugeeei:
```js
// @flow
type Anonymous = interface extends Base {
  m(): boolean;
};
```

`prettier --parser hermes --plugin prettier-plugin-hermes-parser` gives

```js
type Anonymous = interface extends Base {
  m(): boolean,
};
```

A comma. uf prints a semicolon.

## Where the difference is

An `interface` *declaration* separates its members with `;`, and uf is right
there — `flow_types` and `heritage_comments` in the fixture directory both
cover it. An `interface { … }` in a *type* position is an object type, and an
object type's separator is `,`.

Prettier's object printer decides from `node.type === "ObjectTypeAnnotation"`.
uf's `print_object_type` takes an `is_interface` flag from its caller, and the
caller for an interface *type* passes the same `true` as the caller for a
declaration.

## How it was found

Adding `type Anonymous = interface extends Base { … }` to
`heritage_comments.js` while fixing #143. The case is in that fixture with an
empty body and a comment pointing here, so the heritage part is still covered
and the separator is not silently wrong.

## Acceptance

* An `interface { … }` type separates its members with `,` and an `interface`
  declaration with `;`.
* The same for the `declare interface` and `declare class` forms, which are
  declarations either way.
* `heritage_comments.js` gets its members back.

PR #152, opened 2026-09-05
fix(fmt): an interface's `extends` keeps its first target

### LINKED ISSUE #143: fmt: a long multi-target `extends` breaks before the first target, where the hermes plugin does not
Found while fixing #135, and unrelated to it.

```js
interface Several
  extends AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC {
  m(): boolean;
}
```

`prettier --parser hermes --plugin prettier-plugin-hermes-parser` returns that
unchanged. `prettier --parser flow` gives

```js
interface Several
  extends
    AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC {
```

and so does uf.

## Which one uf should match

The hermes one. `crates/uf_fmt/tests/fixtures.rs` says outright that every
expectation in that directory is generated with

```sh
prettier --parser hermes --plugin prettier-plugin-hermes-parser --print-width 100
```

so that is the output "Prettier compatible" is measured against here.

## Where the difference is

Not in the doc that the estree printer builds. `prettier.__debug.printToDoc`
with `--parser flow` gives

```
"interface",
group([" ", "Several", indent(""), indent([
  line, "extends", group(indent([line, "AAAA", ",", line, "BBBB", ",", line, "CCCC"])),
])], { id: heritageGroup })
```

which is what uf builds, byte for byte, and which breaks all three lines
together when it does not fit. The hermes plugin's output cannot come from
that doc, so it overrides the printer somewhere — that is the thing to read
before changing anything here.

Short lists are unaffected: `interface Extended extends Base, Other {` fits,
the group does not break, and every parser agrees. `flow_types` in the fixture
directory covers that case and passes.

## Acceptance

* A long multi-target `extends` matches
  `prettier --parser hermes --plugin prettier-plugin-hermes-parser`.
* The same for `mixins` and `implements` on `declare class`.
* The long case goes back into `heritage_comments.js`, where it is currently
  a comment saying why it is not there.

ubugeeei:
Closes #143. **Stacked on #144** — the diff shows both until that merges.

```js
interface Several
  extends AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC {
```

is what `prettier --parser hermes --plugin prettier-plugin-hermes-parser`
returns for that input. uf put `extends` on a line of its own and the first
target under it.

### Both are Prettier

`prettier --parser flow` agrees with uf. The hermes plugin has its own
`printInterface` that does not: it keeps `"extends "` with its first target and
indents only the ones after it, and it groups on *any* heritage where the
estree printer asks for more than one target or a trailing comment.

The plugin is the one that counts here. `crates/uf_fmt/tests/fixtures.rs` says
every expectation in that directory is generated with it, so it is what
"Prettier compatible" is measured against in this crate.

### The change

The clause is one doc for both counts, differing only in where the indent goes,
and `group_mode` becomes the plugin's condition. The head moves inside the
indent with it — `interface Several` stays whole because the space after
`interface` is a space and not a line, and what the indent reaches is the line
before `extends`.

An `interface { … }` *type* has no name to hang a comment on, so heritage is
the only thing that can make it group; it goes through the same clause now.

### Tests

`heritage_comments.js` gets the long cases back — they were removed with a
comment pointing at this issue — and gains a single long target and an
anonymous interface type. Byte-identical to the plugin.

Not in it: that anonymous type's *members*. uf separates them with `;` where
the plugin uses `,`, which is #151 and nothing to do with heritage. The case is
there with an empty body so the separator is not silently wrong.

```
upstream corpus: 8110 formatted, 10 unparseable, 0 failed
```

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/152"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791206599&installation_model_id=422833&pr_number=152&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F152&signature=b15d49f2f034d97737a0d4edb7e3295d118058a688d02ae715790f2a0d503c6d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/uf_fmt/src/flow/print/types.rs
+++ b/crates/uf_fmt/src/flow/print/types.rs
@@ -276,14 +276,20 @@ impl<'a> Printer<'a> {
                 // An `interface { … }` type has no name, so the only thing
                 // that can make its heritage start a line is having more
                 // than one target.
-                let group_mode = inner.extends.len() > 1;
-                let extends = self.print_interface_extends_list(&inner.extends, key, group_mode);
+                let extends = self.print_interface_extends_list(&inner.extends, key);
                 let body = self.print_object_type(
                     &inner.body.1,
                     NodeRef::ObjectType(&inner.body.0, &inner.body.1),
                     true,
                 );
-                self.concat([self.s("interface"), extends, self.s(" "), body])
+                // An `interface { … }` type has no name to hang a comment
+                // on, so heritage is the only thing that can make it group.
+                let head = if inner.extends.is_empty() {
+                    self.s("interface")
+                } else {
+                    self.group(self.indent(self.concat([self.s("interface"), extends])))
+                };
+                self.concat([head, self.s(" "), body])
             }
             T::Array { inner, .. } => {
                 let argument = self.print_type(&inner.argument);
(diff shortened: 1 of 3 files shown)


ubugeeei:
Merged as part of #156, which squashed the whole stack: `ce9ec6a` on `main` carries this commit, #153's and its own. Nothing here is outstanding, so this is closed rather than left to conflict with a `main` that already has it.

Issue #134, opened 2026-09-05
fmt: a predicate function with no return type loses its colon and the output does not parse


ubugeeei:
A Flow predicate function written without a return type loses the colon, and
the output does not parse.

```js
// @flow
function f(x: mixed): %checks {
  return typeof x === "string";
}
const h = (x: mixed): %checks => typeof x === "string";
```

`uf fmt` gives

```js
function f(x: mixed) %checks {
  return typeof x === "string";
}
const h = (x: mixed) %checks => typeof x === "string";
```

and re-formatting that is

```
syntax error at 2:22: Unexpected token `%`, expected the token `{`
```

## The distinction being missed

Flow has two spellings, and only one of them is handled:

```js
function g(x: mixed): boolean %checks { ... }  // return type + predicate — correct today
function f(x: mixed): %checks { ... }          // inferred predicate — colon dropped
```

The second is the *inferred* predicate: the return type is omitted and the
`%checks` stands where it would have been, so the colon belongs to the
predicate rather than to a type. The printer appears to emit the colon as
part of printing the return type, and there is no return type to print.

`declare function i(x: mixed): boolean %checks(typeof x === "string");` is
also correct today.

## Where it was found

`fbt/packages/babel-plugin-fbt/src/FbtUtil.js:231`, by the formatter corpus
over fbt. It is the only module in the corpus that hits it, and it fails as
"output does not format again" rather than as a layout difference, so it is
the shape of failure that matters most: `uf fmt` writes a file that no
longer builds.

## Acceptance

* Both spellings round-trip, in function declarations, function expressions,
  arrow functions, class methods and object methods.
* `declare function` predicates keep working.
* A fixture pair under `crates/uf_fmt/tests/fixtures`.
* `fbt/packages/babel-plugin-fbt/src/FbtUtil.js` passes
  `upstream_flow_survives_formatting` and comes out of the exclusion list.

Issue #133, opened 2026-09-05
fmt: a right-nested logical chain is laid out differently from a left-nested one


ubugeeei:
**Corrected after checking against `prettier@3.9.6`.** The first version of
this issue said Prettier keeps the parentheses in `a && (b && c)`. It does
not — it drops them, exactly as uf does. The output is Prettier-compatible;
what is wrong is the *layout* of what comes after, and one of the guarantees
that reports it.

## The divergence

```js
// @flow
const b1 = (descriptor instanceof PropertyDescriptor && (descriptor.get === undefined && descriptor.set === undefined));
```

`prettier --parser flow`:

```js
const b1 =
  descriptor instanceof PropertyDescriptor &&
  descriptor.get === undefined &&
  descriptor.set === undefined;
```

`uf fmt`:

```js
const b1 =
  descriptor instanceof PropertyDescriptor &&
  descriptor.get === undefined && descriptor.set === undefined;
```

Both dropped the inner parentheses. Prettier then laid the whole thing out as
one chain of three; uf kept the former right operand as a group of its own,
so it fits on one line and the chain reads as two operands rather than three.

Prettier's `printBinaryishExpressions` flattens a same-operator chain on both
sides before laying it out. uf flattens the left spine only, which is why a
chain written left-nested is already correct:

```js
const a = ((x && y) && z);   // uf and Prettier agree
const b = (x && (y && z));   // uf breaks it differently
```

## Why it also makes the formatter non-idempotent

Pass one prints the flat text above. Pass two parses that text, which is now
a left-nested chain, and lays it out Prettier's way. So `uf fmt` twice is not
`uf fmt` once:

```
- descriptor.get === undefined && descriptor.set === undefined;
+ descriptor.get === undefined &&
+ descriptor.set === undefined;
```

That is the failure the corpus reports for five of the eight modules below.
It is uf's own guarantee, not a Prettier compatibility question.

## The other three are the guarantee being too strict

`support::structure` compares the two trees and sees
`Logical(And, a, Logical(And, b, c))` become
`Logical(And, Logical(And, a, b), c)`, and calls it a changed program.

For `&&`, `||` and `??` it is not one. Both trees evaluate the same operands
in the same order, short-circuit at the same point, and produce the same
value; the re-association is exactly what dropping a redundant parenthesis
means, and Prettier does it too. `+` is a different matter — `(a + b) + c`
and `a + (b + c)` differ for strings and for floats — and Prettier keeps
[...]

PR #170, opened 2026-09-05
test(fmt): an exclusion for a bug that is fixed fails

ubugeeei:
`KNOWN_BROKEN` named nine modules across two bugs. Both bugs are closed —
ubugeeei-prod/uf#133 (a right-nested logical chain re-associating) and
ubugeeei-prod/uf#134 (a predicate function losing its colon) — and all nine
names stayed behind. The modules went on being skipped, the corpus quietly
stopped covering them, and nothing said so.

`no_stale_exclusions` could not have said so. It checked that each excluded
path still matched a file, which is the *other* way a list of excuses rots: a
name that matches nothing. The way that actually bit was a name whose bug was
fixed.

So it now formats what `KNOWN_BROKEN` names, and fails when a module holds all
three guarantees. Putting one of the nine back to check:

```
1 excluded module(s) now hold all three guarantees — take them out of
KNOWN_BROKEN, and close the issue if it was the last one:
  fbt/packages/babel-plugin-fbt/src/FbtUtil.js
```

`KNOWN_SLOW` keeps the existence check and not this one, and the comment says
why: finding out whether a module is still slow means formatting it, and not
formatting it is the entire point of the exclusion.

The three guarantees move into a shared `guarantees()` returning
`Unparseable | Held | Broke(how)`, used by both tests, so the two can never
come to disagree about what "broken" means.

### The corpus, with nothing excluded

```
upstream corpus: 8119 formatted, 1 unparseable, 0 failed
test upstream_flow_survives_formatting ... ok
test no_stale_exclusions ... ok
```

Every Flow module Meta's parser reads across all fifteen repositories — React,
Metro, Relay, React Native, Recoil, Flux, Parcel, Yarn, Prepack, StyleX, fbt,
react-native-web, react-motion, DataLoader and redux-form — comes back
parseable, settled, with the same tree and the same comments. Both exclusion
lists are empty.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/170"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791213651&installation_model_id=422833&pr_number=170&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F170&signature=c915e3f88b1e8b8dc5c4a318006827055eb1aed0d645cd60f99486838bf62f4a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/uf_fmt/tests/upstream_corpus.rs
+++ b/crates/uf_fmt/tests/upstream_corpus.rs
@@ -58,36 +58,29 @@ fn fixtures() -> Vec<String> {
 /// ubugeeei-prod/uf#125 was fixed: `expect.objectContaining` nested
 /// nineteen deep, printed twice per level, so the document was 2^depth.
 ///
-/// Named rather than skipped by a heuristic, so that the day a bug is fixed
-/// this list is what fails and gets deleted. `no_stale_exclusions` keeps it
-/// from rotting the other way.
+/// Named rather than skipped by a heuristic, so that a slow module is a
+/// decision somebody wrote down. `no_stale_exclusions` checks that each name
+/// is still a file in the corpus; it cannot check that the module is still
+/// slow, because finding that out means formatting it, which is the thing
+/// the exclusion exists to avoid.
 const KNOWN_SLOW: [&str; 0] = [];
 
 /// Modules the printer gets wrong, each with the issue that says how.
 ///
 /// Separate from {@link KNOWN_SLOW} because they are different problems and
-/// a single list of excuses hides that. Two bugs, nine modules:
+/// a single list of excuses hides that.
 ///
-/// * **#133** — parentheses around a same-precedence right operand are
-///   dropped, so `a && (b && c)` becomes `a && b && c` and the tree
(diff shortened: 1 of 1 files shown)


ubugeeei:

```
$ git diff --name-only origin/main 170-head -- <the files this branch changes>
(nothing)
```

PR #174, opened 2026-09-05
fix(fmt): an empty block that is not a body keeps its two lines

### LINKED ISSUE #173: fmt: every empty block collapses to {}, where Prettier keeps some open
`uf fmt` collapses every empty block to `{}`. Prettier collapses only the ones
that are a *body*, and keeps a block open when its emptiness is something the
author wrote:

```js
// uf                              // prettier
if (a) {}                          if (a) {
                                   }
if (b) {} else if (c) {            if (b) {
  d();                             } else if (c) {
}                                    d();
                                   }
try {} catch (h) {}                try {
                                   } catch (h) {}
```

`while (f) {}`, `for (;;) {}`, `function g() {}`, `const i = () => {}`,
`class J { k() {} }`, `do {} while (b)` and `component C() {}` are already
right — they collapse in both.

## The rule

Measured against
`prettier --parser hermes --plugin prettier-plugin-hermes-parser
--print-width 100`, block by block. It collapses to `{}` for:

* a function, method, arrow, getter, setter, constructor, generator,
  `component` or `static` block body,
* `for (;;)`, `while` and `do … while`,
* `catch`, **unless** the statement also has a `finally`,
* a `match` arm.

and keeps `{` and `}` on separate lines for `if`, `else`, `try`, `finally`,
`with`, `for … in`, `for … of`, a labelled block, a `case` block, a block on
its own, and a `declare module` / `declare namespace` body.

`for … in` and `for … of` not collapsing where `for (;;)` does looks arbitrary
and is: it is what Prettier does, and the target is byte-identical output.

## Where it was found

The corpus, comparing 500 modules against Prettier. Two of them:

```
prepack/src/utils/reachability.js
react/compiler/…/error.todo-round2_severity_diff.js
```

both `if (…) {}` where Prettier writes the brace on the next line.

## Acceptance

* A fixture pair covering every parent in the list above.
* `components_hooks_match_enums` keeps `_ => {}`.

ubugeeei:
Fixes ubugeeei-prod/uf#173.

`uf fmt` collapsed every empty block to `{}`. Prettier collapses only the ones
that are a *body*, and leaves open a block whose emptiness is something the
author wrote:

```diff
-if (a) {}
+if (a) {
+}
-if (b) {} else if (c) {
+if (b) {
+} else if (c) {
   d();
 }
-try {} catch (h) {}
+try {
+} catch (h) {}
```

### The rule, measured

`{}` for a function, method, arrow, getter, setter, constructor, generator,
`component` or `static` body; for `for (;;)`, `while` and `do … while`; for a
`match` arm; and for a `catch` that is the end of its statement.

`{` and `}` on two lines for `if`, `else`, `try`, `finally`, `with`,
`for … in`, `for … of`, a labelled block, a `case` block, a block on its own,
and a `declare module` / `declare namespace` body.

`for … in` and `for … of` not collapsing where `for (;;)` does looks arbitrary
and is. The list was derived by running each construct through
`prettier --parser hermes --plugin prettier-plugin-hermes-parser
--print-width 100` rather than from memory, because byte-identical output is
the target and taste is not part of it.

### Shape

An `EmptyBlock` argument rather than a rule buried in the block printer. Every
caller but one knows what it is printing and says so; the one exception is a
block *statement*, which is the same node whether it is a loop body, an `if`
branch, a `case` body or a block somebody wrote on its own, so that is the only
place the parent is consulted.

### Tests

`empty_blocks.js` / `.expected.js` — every parent in the list above, one file,
byte-identical to Prettier. It fails on `main` at the first `if`.

The existing `components_hooks_match_enums` fixture caught the `match` arm: an
arm body is a body, not a branch, and Prettier writes `_ => {}`.

Corpus unchanged (`8110 formatted, 0 failed`), and the two modules this was
found in are now byte-identical to Prettier:

```
prepack/src/utils/reachability.js
react/compiler/…/error.todo-round2_severity_diff.js
```

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/174"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791215258&installation_model_id=422833&pr_number=174&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F174&signature=d1345675388b21aa643640aba469e85654ddcc0a3c7c42acbe8ecc5a4d54bc82"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/uf_fmt/src/flow/print/class.rs
+++ b/crates/uf_fmt/src/flow/print/class.rs
@@ -5,9 +5,9 @@ use uf_flow::Loc;
 use uf_flow::ast::{class, expression, function, statement, types};
 
 use super::Printer;
-use super::statement::EmptyBlock;
 use super::assignment::Rhs;
 use super::parens::is_member;
+use super::statement::EmptyBlock;
 use crate::doc::{Doc, HARDLINE, LINE, SOFTLINE};
 use crate::flow::comments::{Marker, Placement};
 use crate::flow::node::{Class, NodeKey, NodeRef};
--- a/crates/uf_fmt/src/flow/print/class.rs
+++ b/crates/uf_fmt/src/flow/print/class.rs
@@ -5,6 +5,7 @@ use uf_flow::Loc;
 use uf_flow::ast::{class, expression, function, statement, types};
 
 use super::Printer;
+use super::statement::EmptyBlock;
 use super::assignment::Rhs;
 use super::parens::is_member;
 use crate::doc::{Doc, HARDLINE, LINE, SOFTLINE};
@@ -360,7 +361,7 @@ impl<'a> Printer<'a> {
             class::BodyElement::PrivateField(field) => p.print_private_field(field),
             class::BodyElement::StaticBlock(block) => {
                 let key = NodeRef::ClassMember(member).key();
-                let body = p.print_block_body(&block.body, key);
+                let body = p.print_block_body(&block.body, key, EmptyBlock::Collapsed);
                 p.concat([p.s("static "), body])
             }
             class::BodyElement::DeclareMethod(method) => {
--- a/crates/uf_fmt/src/flow/print/function.rs
+++ b/crates/uf_fmt/src/flow/print/function.rs
@@ -11,6 +11,7 @@ use uf_flow::Loc;
 use uf_flow::ast::{expression, function, pattern, statement, types};
 
 use super::Printer;
+use super::statement::EmptyBlock;
 use super::assignment::{Layout, PrintArgs};
 use super::call::{is_test_call, parameter_count};
 use super::parens::{is_binaryish, is_call_like, is_jsx, starts_with_no_lookahead_token};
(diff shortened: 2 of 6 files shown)


ubugeeei:

```
$ git diff --name-only origin/main 174-head -- <the files this branch changes>
(nothing)
```

Issue #162, opened 2026-09-05
lsp: `uf lsp` reads stdin to EOF, so it never answers an editor


ubugeeei:
`uf lsp` answers one request and exits. Against a real editor it answers
nothing at all.

```rust
let mut input = String::new();
std::io::stdin().read_to_string(&mut input)?;
```

`read_to_string` returns at EOF. An editor holds the pipe open for the life of
the session, so `uf lsp` blocks there forever, never reads the `initialize` it
was sent, and never writes a byte. The editor times out.

`cli.rs`'s `lsp_initialize_returns_native_capabilities` passes because
`write_stdin` closes the pipe — which is the one thing an editor never does.

## And the capabilities are not backed

The response advertises:

```json
"documentFormattingProvider": true,
"diagnosticProvider": { "interFileDependencies": true, "workspaceDiagnostics": true }
```

Nothing serves any of them. There is no `textDocument/formatting` handler, no
document store, and no diagnostics.

`editors/` holds seven integration directories — VS Code, Neovim, Emacs, Vim,
Helix, Zed, Cursor — - built on this.

## What it needs

A message loop, which is most of the gap:

* read `Content-Length` framing until `exit`, rather than to EOF;
* `initialize` / `initialized` / `shutdown` / `exit`;
* `textDocument/didOpen`, `didChange` (full sync, which is what
  `textDocumentSync: 1` already claims) and `didClose`, kept in a document
  store;
* `textDocument/formatting` answered from `uf_fmt::format_source` — one
  `TextEdit` replacing the whole document, which is what a printer that
  reprints from the tree produces;
* capabilities reduced to what is actually served, so an editor does not ask
  for diagnostics that never arrive.

## Acceptance

* A test that speaks the protocol the way an editor does: two framed messages
  on one open pipe, both answered, `exit` ending the loop. The current test
  passes against a server that cannot work.
* `uf fmt` and `uf lsp` produce the same bytes for the same document.
* Advertised capabilities and implemented handlers are the same list.

PR #190, opened 2026-09-05
feat(lsp): an editor is told what is wrong with the file

### LINKED ISSUE #189: lsp: an editor gets formatting and no diagnostics
`uf lsp` serves formatting and nothing else. An editor connected to it gets
`uf fmt` on save and no idea that a file does not parse — while `uf lint`,
run in a terminal beside it, says exactly what is wrong and where.

```
$ uf lint
  warning[react/no-default-export-component]: framework routes are wired by
  name; export components with a named export
   --> app/_uf.page.js:23:1
```

The same diagnostics belong in the editor. `uf_lint::lint_source` already
takes one file and returns them, and `flow/syntax` already turns the parser's
own errors into diagnostics — which is the one an editor most wants while a
file is still being typed.

## What it needs

* Publish on `didOpen` and on every `didChange`, and clear on `didClose`: an
  editor keeps whatever it was last told, so a file that is fixed and one that
  is closed both have to be said out loud.
* `textDocument/publishDiagnostics` is a *notification*, so nothing is
  advertised in `initialize`. The pull-model `diagnosticProvider` that used to
  be advertised and served nothing stays gone — that was
  ubugeeei-prod/uf#162.
* Two conversions the protocol needs and the linter does not do:
  * the linter's line is 1-based and LSP's is 0-based;
  * the linter's column is a **byte** offset and LSP's is a **UTF-16 code
    unit** offset. They agree only while the line is ASCII, so `const π = 1;`
    would put every marker on that line one column to the right.
* The URI is a `file://` URI and the linter is path-sensitive — `flow/syntax`
  claims `.js`, `.jsx`, `.mjs` and `.cjs` and nothing else — so it has to be
  decoded to a path, percent escapes included.
* The underline should be the span `uf lint` underlines, so the marker in the
  editor and the caret in the terminal cover the same text.

## Acceptance

* Opening a file with a syntax error publishes `flow/syntax` at severity 1.
* Opening a file that trips a rule publishes it at the rule's severity.
* Changing to a clean document publishes an empty list; closing does too.
* Byte-to-UTF-16 and URI-to-path pinned by unit tests, non-ASCII included.

ubugeeei:
Fixes ubugeeei-prod/uf#189.

`uf lsp` served formatting and nothing else, so an editor connected to it got
`uf fmt` on save and no idea that a file did not parse — while `uf lint`, in a
terminal beside it, said exactly what was wrong and where.

It now publishes diagnostics from the same `uf_lint::lint_source` that
`uf lint` calls, on open and on every change, so a marker in the editor and a
line in the terminal are the same diagnostic. Driven against the real binary:

```json
{"method":"textDocument/publishDiagnostics","params":{
  "uri":"file:///a.js",
  "diagnostics":[{
    "range":{"start":{"line":1,"character":10},"end":{"line":1,"character":11}},
    "severity":1, "source":"uf", "code":"flow/syntax",
    "message":"Unexpected token `;`"}]}}
```

```json
{"...":"a rule, on a scaffolded page",
  "range":{"start":{"line":1,"character":0},"end":{"line":1,"character":6}},
  "severity":2, "code":"react/no-default-export-component"}
```

Closing publishes an empty list rather than falling silent: an editor keeps
whatever it was last told, so a closed file would keep its markers.

### Nothing new is advertised

`publishDiagnostics` is a notification the server sends, not a request the
editor makes. The pull-model `diagnosticProvider` that used to be advertised
and served nothing — ubugeeei-prod/uf#162 — stays gone, and the test that says
so is still there with the reason updated.

### Two conversions the protocol needs

* Lines are 1-based in the linter and 0-based in the protocol.
* Columns are **byte** offsets in the linter and **UTF-16 code units** in the
  protocol. They agree only while the line is ASCII: `const π = 1;` would put
  every marker on that line one column right, and `const 🦀 = 1;` two.
  `a_byte_column_becomes_a_utf16_character` pins both, plus a column that
  lands inside a character and one past the end of the line.

And the URI is decoded to a path, percent escapes included, because the linter
is path-sensitive — `flow/syntax` claims `.js`, `.jsx`, `.mjs` and `.cjs`, so a
project under `My Project` arriving as `My%20Project` would be a file silently
not parsed.

The underline is the span `uf lint` underlines — `identifier_span`, the same
helper — so the marker and the caret cover the same text.

### Tests

* `lsp_publishes_diagnostics_when_a_document_opens` — a syntax error comes back
  as `flow/syntax` at severity 1, zero-based, with no id to answer.
* `lsp_republishes_on_change_and_clears_on_close` — three publishes, two of them
  empty.
* Unit tests for both conversions.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/190"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791218690&installation_model_id=422833&pr_number=190&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F190&signature=c42842d513e81c6fcef23569eb928a7fdcd3bb66a6f5737d10d26c444665f526"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/uf_cli/src/commands/dev.rs
+++ b/crates/uf_cli/src/commands/dev.rs
@@ -11,11 +11,12 @@ use std::io::{BufRead, IsTerminal, Write};
 use anyhow::{Context, Result, bail};
 use camino::Utf8Path;
 use serde_json::{Value, json};
-use uf_config::{FmtConfig, load_config};
+use uf_config::{FmtConfig, UniflowedConfig, load_config};
 use uf_infra::FxHashMap;
 use uf_router::write_router_manifest;
 use uf_term::{KeyValue, Status, Tone};
 
+use crate::commands::lint::identifier_span;
 use crate::commands::vite::{Driver, Event, package_dir, render_error, render_log, resolve_host};
 use crate::support::{plural, project_label};
 use crate::ui::Ui;
@@ -145,9 +146,11 @@ pub(crate) fn dev(cwd: &Utf8Path, ui: &mut Ui, args: DevArgs) -> Result<()> {
 /// `TextEdit` over the whole file, which is what a printer that reprints from
 /// the syntax tree produces.
 ///
-/// Diagnostics are *not* advertised. They were before, and nothing served
-/// them; an editor that asks for what it is offered and receives nothing is
-/// worse off than one that was never offered it.
+/// Diagnostics, from the same `uf_lint::lint_source` that `uf lint` calls,
+/// pushed on open and on every change. `publishDiagnostics` is a
+/// notification the server sends rather than a request the editor makes, so
+/// there is no capability to advertise — the pull-model `diagnosticProvider`
+/// that used to be advertised, and served nothing, stays gone.
 pub(crate) fn lsp() -> Result<()> {
     let stdin = std::io::stdin();
     let mut stdout = std::io::stdout().lock();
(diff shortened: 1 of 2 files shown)


ubugeeei:

```
$ git diff --name-only origin/main 190-head -- <the files this branch changes>
(nothing)
```

PR #192, opened 2026-09-05
fix(installer): the download goes where TMPDIR says

### LINKED ISSUE #191: installer: the download ignores TMPDIR and fails where the system temp is not writable
`curl -fsSL https://setup.uniflowed.dev | sh` fails on a machine whose system
temp directory is not writable, and fails like this:

```
  target   aarch64-apple-darwin
  version  0.0.0-alpha.2  (prerelease — no stable release yet)

mktemp: mkdtemp failed on /var/folders/…/tmp.WC7zFqHuKe: Operation not permitted
```

which names neither the cause nor the fix. Setting `TMPDIR` does not help:

```sh
tmp_dir="$(mktemp -d)"    # install.sh:532
```

BSD `mktemp -d` given no template ignores `TMPDIR` and uses the system
directory. The same trap is already written down two directories away, in
`tools/release/verify-npm.sh`:

```sh
# With a template: BSD `mktemp -d` ignores `TMPDIR` without one, and this
# script is run by hand on macOS as often as by CI on Linux.
work="$(mktemp -d "${TMPDIR:-/tmp}/uf-verify-npm.XXXXXX")"
```

## Who hits it

Anyone whose system temp is unusable and who set `TMPDIR` expecting it to be
honoured: a sandbox, a container with a read-only `/tmp`, a locked-down CI
image, a shared host with a full temp filesystem. The installer is the first
thing a user runs, and this is a failure before a single byte is downloaded.

`tools/release/test-install.sh` has its own `mktemp -d` with the same shape.

## Acceptance

* `TMPDIR=/somewhere/writable curl … | sh` downloads into that directory.
* A `TMPDIR` that does not exist stops the install rather than silently using
  the system temp — which is also the case that tells the two behaviours
  apart, and belongs in `test-install.sh`.

ubugeeei:
Fixes ubugeeei-prod/uf#191.

`curl -fsSL https://setup.uniflowed.dev | sh` fails on a machine whose system
temp directory is not writable, and fails like this:

```
  target   aarch64-apple-darwin
  version  0.0.0-alpha.2  (prerelease — no stable release yet)

mktemp: mkdtemp failed on /var/folders/…/tmp.WC7zFqHuKe: Operation not permitted
```

Setting `TMPDIR` does not help, because BSD `mktemp -d` given no template
ignores it and uses the system directory anyway.

```diff
-tmp_dir="$(mktemp -d)"
+tmp_dir="$(mktemp -d "${TMPDIR:-/tmp}/uf-install.XXXXXX")"
```

The same trap is already written down two directories away, in
`tools/release/verify-npm.sh`, which is where it was recognised.

### Evidence

With the template, the **published** `uf@0.0.0-alpha.2` installs from
setup.uniflowed.dev into a sandbox with no writable system temp:

```
  ✓ downloaded uf-aarch64-apple-darwin.tar.gz
  ✓ verified   sha256 362036f709d7
  ✓ unpacked   …/runtimes/uf@0.0.0-alpha.2
  ✓ linked     uf, ufr, ufx

  uf 0.0.0-alpha.2 is ready.
```

and the binary it linked runs:

```
$ uf --version
uf_cli 0.0.0-alpha.2
$ uf create app react demo
✓ created 8 files
```

### Test

`test-install.sh` grows the case that tells the two behaviours apart: pointed
at a `TMPDIR` that does not exist, an installer that honours it stops, and one
that ignores it carries on using the system temp. It checks the ordinary
direction too — a `TMPDIR` that does exist installs — because a guard that
only asserts a failure is a guard that passes when nothing works.

`test-install.sh`'s own `mktemp -d` had the same shape and is fixed with it.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/192"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791219091&installation_model_id=422833&pr_number=192&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F192&signature=c121b61118292bb2edc6be2abf4cc8c3dd5c73e7721e249f06fa0c573a9936be"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/infra/cloudflare/setup-assets/install.sh
+++ b/infra/cloudflare/setup-assets/install.sh
@@ -529,7 +529,14 @@ archive="uf-${target}.tar.gz"
 archive_url="${channel_url}/${archive}"
 checksum_url="${archive_url}.sha256"
 
-tmp_dir="$(mktemp -d)"
+# With a template, so `TMPDIR` is honoured. BSD `mktemp -d` given no template
+# ignores it and uses the system directory, which is not where a reader who
+# set `TMPDIR` asked for their downloads to land — a restricted CI image, a
+# sandbox, a container whose default temp is read-only. The same trap is
+# written down in `tools/release/verify-npm.sh`, which is where this was
+# noticed: an installer that fails here fails with `mkdtemp failed`, which
+# names neither the cause nor the fix.
+tmp_dir="$(mktemp -d "${TMPDIR:-/tmp}/uf-install.XXXXXX")"
 cleanup() {
   rm -rf "$tmp_dir"
 }
--- a/tools/release/test-install.sh
+++ b/tools/release/test-install.sh
@@ -51,7 +51,8 @@ target="$(rustc -vV | awk -F': ' '/^host:/ { print $2 }')"
 archive="uf-${target}.tar.gz"
 [ -f "${release_dir}/${archive}" ] || fail "missing ${release_dir}/${archive}"
 
-work="$(mktemp -d)"
+# See `install.sh`: without a template, BSD `mktemp -d` ignores `TMPDIR`.
+work="$(mktemp -d "${TMPDIR:-/tmp}/uf-test-install.XXXXXX")"
 server_pid=""
 cleanup() {
   [ -n "$server_pid" ] && kill "$server_pid" 2>/dev/null || true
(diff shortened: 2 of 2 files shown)


ubugeeei:

```
$ git diff --name-only origin/main 192-head -- <the files this branch changes>
(nothing)
```

Issue #130, opened 2026-09-05
release: alpha.3 is blocked on npm trust bindings


ubugeeei:
`tools/release/preflight.sh` says a `uf@*` tag would half-send the release
today. Ten of the fifteen names in `published-packages.txt` are not on the
registry, and each has to be bound to this repository's publish workflow
before a tag can send it:

```
Not on the registry yet: fetch graphql hooks host query react-testing relay
server stylex ui
```

Five are already there: `core`, `router`, `test`, `vite`, and one more.

## Why this is a blocker rather than a task

The binding is made **as ubugeeei, not as the workflow**, so it is the one
step in the release that nobody else and no agent can do:

```sh
npm login
tools/release/trust-npm.sh
```

`trust-npm.sh` is idempotent — a name already bound is reported and left
alone — and it binds names that have never been published, which is what
lets the first release go out over OIDC like every one after it.

## What happens if a tag goes out first

`publish.yml` publishes the list in order. The bound names go out, the first
unbound one fails the job, and the release is half-sent. It is recoverable —
npm is additive, so re-running publishes the rest — but the version is then
partly public and partly not, which is worse than not having tagged.

## State

`main` is green and has five merged changes since `uf@0.0.0-alpha.2`
(#122, #123, #124, #127, #129). The next tag is `uf@0.0.0-alpha.3`, per the
`v0.0.0-alpha.N` prerelease series in the redundancy guide.

Recording this rather than tagging, per the guide: *"When one task is
blocked, record the blocker and continue independent work without pretending
the blocked capability is finished."*

ubugeeei:
Two names have joined the list since this was filed, and both are load-bearing
rather than aspirational:

* **`core`** — `@uniflowed/stylex` and `@uniflowed/react-testing` declare it,
  and it is where `nativeRuntimeRequired` lives. Publishing either without it
  produces a package that resolves to nothing. Found by the closure check in
  #147.
* **`validator`** — `@uniflowed/ui` imports it for `Schema` and now declares
  it. Found by `every_uniflowed_import_is_declared` in #150.

So the list is:

```
core fetch graphql hooks host query react-testing relay server stylex ui validator
```

Twelve names, one command, and it is still the one step nobody else and no
agent can run:

```sh
npm login
tools/release/trust-npm.sh
```

## What has changed around it

**The half-send is now caught before it can happen.** #147 added
`tools/release/verify-npm.sh --closure-only`, which `ci` runs on every push and
[...]

PR #194, opened 2026-09-05
fix(release): ask npm what it can do, not what number it is

ubugeeei:
Run against a real npm, the guard added in ubugeeei-prod/uf#188 fired and
reported only this:

```
trust-npm: npm rejected the arguments this script passes:
npm warn "publish.yml" is being parsed as a normal command line argument.
npm warn "ubugeeei-prod/uf" is being parsed as a normal command line argument.
npm warn Unknown cli config "--file".
npm warn Unknown cli config "--repository".
```

Two things were wrong.

### The version gate is too loose

`npm trust github` and its `--file` and `--repository` arrived partway through
npm 11; the gate only asked for npm 11. And an npm that has `trust` without
those arguments **does not fail on them** — npm warns about a flag it does not
know and carries on with the value as a positional. So the run looks like it
worked and binds nothing.

It is a feature probe now: `npm trust github --help` has to mention
`--repository`. It reports the version rather than leaving it to be guessed:

```
trust-npm: this npm does not have `npm trust github --file --repository`.

  npm --version   11.6.2

That subcommand is how a package name is bound to this repository's publish
workflow, and it is the whole of the release bootstrap. Upgrade npm and run
this again:

  npm install -g npm@latest
```

### The guard threw away stdout

`>/dev/null 2>"$errors"` kept the warnings and discarded the error underneath
them, which is why the report above is three warnings and no reason. Both
streams are kept now, and npm's version is printed after them, because that is
what the next question is going to be.

### Checked

Against a stub npm both ways: one that answers `--help` without `--repository`
stops with the upgrade message and binds nothing; one that answers with it
reaches the loop and binds all seventeen names.

Context in ubugeeei-prod/uf#130.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/194"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791220153&installation_model_id=422833&pr_number=194&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F194&signature=5e8180522e89fd6afe721c58bd34d5b9b5e18031842e898e2fe42b2046b9c6c4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -276,6 +276,11 @@ jobs:
       # resolves to nothing. No network, so it belongs here rather than in the
       # pre-tag check alone.
       - run: ./target/release/uf run release:closure
+      # And that the one command the release bootstrap depends on still runs.
+      # It is invoked by hand, once, by one person, so a mistake in it costs a
+      # round trip and blocks a release — it has cost three. Stubs, so this
+      # needs no npm session and no network.
+      - run: ./target/release/uf run release:trust:test
 
   upstream-integrations:
     name: Upstream Integrations
--- a/tools/release/bootstrap-publish.sh
+++ b/tools/release/bootstrap-publish.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env sh
+# Publish, once, the names the registry does not have yet.
+#
+# `npm trust` binds a name the registry already has. It cannot create one:
+#
+#   npm error code E404
+#   npm error 404 Not Found - POST .../@uniflowed%2fcore/trust
+#
+# So the first publish of a *new* name cannot go over OIDC, because there is
+# nothing to bind the workflow to. It has to come from a person, once, and
+# then `tools/release/trust-npm.sh` binds it and every release after that is
+# the workflow's. `uf.config.js` calls this `publish.firstPublish.localBootstrap`.
+#
+#   npm login
+#   tools/release/bootstrap-publish.sh
+#
+# It publishes only what is missing. A name that is already on the registry is
+# left alone whatever version it is at — moving an existing package to a new
+# version is the release workflow's job, not this one's.
+#
+# Every publish is shown before anything is sent, and nothing is sent without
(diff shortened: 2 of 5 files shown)


ubugeeei:
Reopening my own conclusion: removing `--allow-publish` in ubugeeei-prod/uf#188
was wrong. It was right for the npm in front of me and wrong for the npm this
is actually run on, which requires it:

```
npm error At least one permission flag is required
          (--allow-publish, --allow-stage-publish)
```

`npm trust github`'s flags have moved in **both** directions across npm
releases — required, then unknown, then required again — so neither hardcoding
is correct and the version number is not the question to ask. The help is:

```sh
permission=""
if npm trust github --help 2>&1 | grep -q -- '--allow-publish'; then
  permission="--allow-publish"
fi
```

The argument list is one function now, shared by the dry run and by every real
bind. That is the actual lesson from #188: the guard passed while the bind it
was guarding could not run, because they were two separate argument lists.

Checked against three stub npms — requires the flag, rejects the flag, too old
for the subcommand — binding 17, 17 and 0 respectively.

ubugeeei:

```
$ git diff --name-only origin/main 194-head -- <the files this branch changes>
(nothing)
```

Issue #441, opened 2026-09-06
fmt: `uf fmt` fails in a freshly scaffolded project, over the manifest `uf create` wrote


ubugeeei:
`uf create app react` scaffolds a project. `uf fmt` in that project fails:

```
! biome is not installed, so 1 non-Flow files were left alone. Install it, or set `fmt.nonFlow.formatter` to "none" in uf.config.js.

! 0 files of 7 need formatting
error: biome is not installed, so 1 non-Flow files were left alone. Install it, or set `fmt.nonFlow.formatter` to "none" in uf.config.js.
```

Nothing needs formatting. The one non-Flow file is `package.json` — the manifest
uf itself just wrote. `fmt.nonFlow.formatter` defaults to `biome`, and
`uf create` does not add Biome to the project, so the default configuration and
the default scaffold disagree on the first command after `uf install`.

Reproduced on the published `0.0.0-alpha.7` binary and on a `main` build, from a
clean scaffold, so it is not a version skew.

## Why it matters more than it looks

`uf fmt` is one of the five commands the docs tell a new reader to try, and the
one whose whole promise is that it needs no setup. A failure here is the first
thing uf does wrong to somebody who has done nothing wrong, and the failure is
about a file they did not write.

## The shape of the answer is a decision

The message is good — it names the cause and two fixes. The question is which of
these uf should do, and it is a product decision rather than a bug fix:

1. **The scaffold installs Biome.** Honest, and it adds a dependency to every
   new project for one file.
2. **The scaffold sets `fmt.nonFlow.formatter: "none"`.** No dependency, but
   then `uf fmt` silently does nothing to `package.json` and a project that
   later adds CSS gets no formatting for it either, with nothing saying so.
3. **A missing non-Flow formatter is a warning, not an error.** `uf fmt` did
   everything it could do and reports honestly what it skipped; the exit code
   says whether the files uf *can* format are formatted. This is what
   `uf fmt --check` in CI would want too.
4. **uf formats JSON itself.** `package.json` is the only non-Flow file most
   projects have at scaffold time, and JSON is small.

There is also a grammar bug in the message either way: "1 non-Flow **files**".

ubugeeei:
Correction to the issue as filed: the non-Flow file is **`package-lock.json`**, not `package.json`. `package.json` has been exempt from the non-Flow formatter since #112 — which is why the reproduction needs all three commands rather than two: the file only appears after `uf install`.

Found while fixing it in #446.

Issue #420, opened 2026-09-06
test: five `vite.rs` cases panic under `cargo test --workspace` and pass when that target runs alone


ubugeeei:
`cargo test --workspace` cannot pass in a sandbox that refuses `TcpListener::bind`,
and it does not have to: `loopback_ready()` in `crates/uf_cli/tests/vite.rs`
already exists to detect that, and its own comment names `UF_ALLOW_FIXTURE_SKIP=1`
as the opt-out. Five cases panic instead:

```
this test needs a loopback socket and could not bind one:
Operation not permitted (os error 1)
   at crates/uf_cli/tests/vite.rs:936
```

Two separate agents hit it on this machine today, and one established the part
that makes it a defect rather than an environment note: **the same target run on
its own passes 13 of 13.** It fails only when several test binaries run at once.
So this is not "the sandbox forbids binding" — the sandbox permits it, up to
some concurrent limit, and the guard reports the refusal as a hard panic rather
than as the skip it was written for.

That means a contributor sees five failures whose message says the environment
is at fault, on a machine where the feature works, and the advice in the comment
(`UF_ALLOW_FIXTURE_SKIP=1`) silently skips real coverage rather than fixing
anything.

## What it needs

- decide what `loopback_ready()` means when `bind` fails *transiently*: a retry
  with backoff distinguishes "this machine cannot bind" from "not right now",
  and only the first is worth skipping for;
- the panic message should say which of the two it concluded and how it decided,
  since the current one asserts the stronger claim;
- and whatever it does, `cargo test --workspace` and `cargo test -p uf_cli --test
  vite` should not disagree about whether the feature works.

CI binds sockets and runs all thirteen, so this is about the local run — which is
the one every contributor does first.

Issue #408, opened 2026-09-06
release: `uf create` pins `latest`, and `latest` on npm is five releases behind


ubugeeei:
`uf create app react` writes this manifest (`crates/uf_project/src/template.rs:80`):

```json
"dependencies": {
  "@uniflowed/config": "latest",
  "@uniflowed/react": "latest",
  "@uniflowed/router": "latest",
  "@uniflowed/vite": "latest"
},
"devDependencies": { "@uniflowed/test": "latest" }
```

`latest` is a dist-tag, and on npm it does not point at this release. Read from
the registry today, with `uf@0.0.0-alpha.7` the current release:

| package | `latest` | `alpha` |
|---|---|---|
| @uniflowed/config | **0.0.0-alpha.1** | 0.0.0-alpha.7 |
| @uniflowed/react | **0.0.0-alpha.1** | 0.0.0-alpha.7 |
| @uniflowed/router | **0.0.0-alpha.1** | 0.0.0-alpha.7 |
| @uniflowed/vite | **0.0.0-alpha.1** | 0.0.0-alpha.7 |
| @uniflowed/test | **0.0.0-alpha.1** | 0.0.0-alpha.7 |
| @uniflowed/host | **0.0.0-alpha.2** | 0.0.0-alpha.7 |
| @uniflowed/ui | **0.0.0-alpha.2** | 0.0.0-alpha.7 |

All seventeen published names are like this: `latest` is alpha.1 or alpha.2,
`alpha` is alpha.7.

So the first thing a new person does — install uf, run `uf create app react`,
install — gives them a project pinned to packages five releases old, and not
even consistently: `@uniflowed/react` at alpha.1 beside a transitive
`@uniflowed/host` at alpha.2. The binary that scaffolded it is alpha.7.

Reproduce without installing anything:

```
curl -s https://registry.npmjs.org/@uniflowed%2freact | python3 -c 'import json,sys; print(json.load(sys.stdin)["dist-tags"])'
{'latest': '0.0.0-alpha.1', 'alpha': '0.0.0-alpha.7'}
```

## Why

`.github/workflows/publish.yml:80` chooses the dist-tag from the version:

```sh
tag=latest
case "$version" in
  *-alpha*) tag=alpha ;;
  *-beta*) tag=beta ;;
  *-rc*) tag=rc ;;
esac
```

That is the right rule for a package that has a stable release to protect —
a prerelease must not displace it. These packages have no stable release. The
[...]

Issue #252, opened 2026-09-06
rsc: the analysis is complete and nothing reads it — every page ships to the browser and no server action is callable


ubugeeei:
`crates/uf_rsc` is 5,100 lines of finished analysis: directive scanning, a resolved module graph with reachability and client boundaries, an HMAC-keyed action registry with a documented threat model, a manifest, and generated Flow types for every action. Nothing downstream reads any of it. The bundle has no server/client split, and no server action can be called.

`crates/uf_rsc/src/types.rs:48-52` says the second half out loud:

> * **A dispatcher.** There is no runtime that takes a name and calls an action, so this file declares no function.

## What Next.js does

A module marked `"use server"` has its exports replaced, in the client bundle, by references that POST to the server; React serializes the arguments, the server looks the id up, runs the function and streams the result back, and `<form action={fn}>`, `useActionState` and `useFormStatus` are built on that. A module marked `"use client"` becomes a boundary: its code goes to the browser and everything above it does not. `next build` fails when a Client Component imports server-only code, and names the import chain.

## What uf does today, measured on its own documentation site

`docs/dist/docs/uf-rsc-manifest.json`, from the checked-in build:

```json
{ "modules": 12, "clientBoundaries": 0, "clientBundleRoots": 0, "serverActions": 0, "diagnostics": 5 }
```

Zero client boundaries — correct, because `grep -rn "use client" docs/app/` finds nothing. And `ls docs/dist/docs/assets/ | grep -c '^_uf.page-.*\.js$'` is **17**: every page module in the site is a separate chunk in the *client* bundle. `docs/architecture.md:381` says "Server Components are the default"; what the build produces is ordinary SSR with full hydration, and the server/client distinction exists only in a JSON file beside it.

The reason is `packages/vite/internal/routes.js:200` — `routesModuleSource` emits `page: () => import(<file>)` for every route into `virtual:uf/routes`, and `virtual:uf/client` (`routes.js:255`) imports that same table. One table, both bundles. `packages/vite/index.js` registers two plugins, `uf:flow` (line 123) and `uf:mdx` (line 313); there is no `uf:rsc`. `crates/uf_transform` never sees a directive: grepping it for `use server` finds one string, in a test fixture at `crates/uf_transform/src/babel.rs:606`, asserting that a directive prologue survives the transform unchanged.

So the comment in `packages/server/index.js:12-14` —

[...]

PR #442, opened 2026-09-06
docs: a README for somebody who has never heard of uf

### LINKED ISSUE #435: docs: the README says uf is not usable and nothing is published, seven releases later
`README.md` is the first thing anyone reads about uf, and its warning block is
the first thing in it:

> **🚧 Under active construction — not usable yet. 🚧**
>
> - **Nothing is published.** There is no release on GitHub, nothing on
>   crates.io, and nothing on npm. `curl … | sh` will not install anything until
>   the first tag exists.

Every sentence of that is now false, and the "Usage" section repeats it:

> Once a release exists, this is how `uf` will be installed. **Neither command
> works yet** — no tag has been cut, so there is nothing behind either URL.

Checked just now:

```
$ curl -fsSL https://setup.uniflowed.dev | sh
  target   aarch64-apple-darwin
  version  0.0.0-alpha.7  (prerelease — no stable release yet)
  ✓ downloaded uf-aarch64-apple-darwin.tar.gz
  ✓ verified   sha256 ab8c17c6adc1
  ✓ unpacked   …/runtimes/uf@0.0.0-alpha.7
  ✓ linked     uf, ufr, ufx into …/bin
  uf 0.0.0-alpha.7 is ready.
```

There are **eight** GitHub releases, `uf@0.0.0-alpha.0` through
`uf@0.0.0-alpha.7`, fourteen binary assets each, and **seventeen**
`@uniflowed/*` packages on npm at the `alpha` tag. So a person who lands on the
repository is told the tool does not exist, on the same day the tool installs
itself in four seconds and scaffolds a working project.

## The rest of what is stale

- **"The build is being moved onto Vite. The dev server and bundler in this
  repository are being deleted as Vite takes over, so `uf dev` and `uf build`
  are mid-migration."** That migration is done. There is no bundler crate and no
  dev-server crate; `crates/uf_bundle` measures bundle size and enforces
  budgets, which is a different job that its own module docs describe.
- **"Code inside a tagged template — GraphQL, CSS — is left as written, where
  Prettier reformats it."** `crates/uf_fmt/src/graphql/` is a lexer, a parser
  and a printer, with fixtures. GraphQL inside a tagged template *is* formatted
  (#238).
- **"Until then, build from a checkout"** is offered as the only way to get uf.
  It is now the contributor's path, not the reader's.

## What it should be instead

Not a smaller warning — a README that does the job the user asked docs to do:
written for somebody who has never heard of uf, saying what it is, what it is
for, what it does and does not do today, and how to start. The "Defaults"
section is a 60-item list of everything uf intends to be, which is a roadmap
wearing a feature list's clothes; a reader cannot tell from it which of those
sixty things they can use this afternoon.

`docs/roadmap.md` and the docs site already keep Implemented / Experimental /
Planned apart. The README should point at them rather than restate them badly,
and should be honest that this is `0.0.0-alpha`: interfaces move, and the
guide says where uf loses to the tools it means to replace.

ubugeeei:
`README.md` opened by telling the reader uf was "not usable yet" and that
"Nothing is published. There is no release on GitHub, nothing on crates.io, and
nothing on npm." Eight releases and seventeen npm packages later, that was the
first thing anyone who found the repository read, and the Usage section repeated
it: "**Neither command works yet** — no tag has been cut, so there is nothing
behind either URL."

The Vite migration it called "mid-migration" is finished — there is no bundler
crate and no dev-server crate — and the paragraph saying code inside a tagged
template is "left as written" describes a formatter that has had a GraphQL
lexer, parser and printer since #238.

Rewritten as a front door rather than a construction sign.

## What it is now

What uf is and who it is for, in language that assumes nothing: React is written
in Flow, Flow has the React syntax, and using that syntax means making five
tools agree about it. Then what `0.0.0-alpha` actually means, with every limit
linked to its issue; an install and a first project with output from a run
rather than from memory; a command table; the one config file; and a map of the
documentation.

## What went, and why

- **"Defaults", sixty bullets of intent presented as a feature list.** A reader
  could not tell which of the sixty they could use this afternoon. Implemented,
  experimental and planned are three named states now, each pointing at the page
  that keeps its list — `/reference/cli` and `/reference/config` describe only
  what exists, `/guide/scope` gives every gap an issue number, and
  `docs/roadmap.md` sorts them. A fourth copy of a list is a fourth thing to
  keep true.
- **The forty-four `@uniflowed/*` names.** Names without descriptions teach
  nothing; `/reference/packages` annotates every one and says which are
  published.
- **"Current Execution Surface", twelve bullets of internal artefacts**, four of
  them stale: `uf install` runs the project's package manager rather than only
  writing `uf.lock`, and `uf upgrade` does not do what it said.

## Running the quickstart to check it

The quickstart was run rather than assumed, and it does not fully work on the
published release. Three things a reader hits today, so the README names them
with their issues rather than letting them be discovered three commands in. The
block is scoped to a version and says so.

1. `uf create` on the published binary writes `"latest"`, and npm still resolves
   that to `0.0.0-alpha.1`, so a new project installs alpha.1 packages against
   an alpha.7 binary and `uf test` stops with `@uniflowed/host is not installed`
   (#408, fixed by #416, unreleased at the time of writing).
2. Pinning to `alpha` instead trades that for `ERR_MODULE_NOT_FOUND`:
   `@uniflowed/host` names `./write-atomically` in `exports` and leaves it out
   of `files`, so the tarball lacks it — checked against the alpha.5, alpha.6
   and alpha.7 tarballs (#410, fixed on `main`, unreleased at the time of
   writing).
3. `uf fmt` exits 1 in a scaffolded project, because Biome is the default
   non-Flow formatter, `package.json` is a non-Flow file, and the scaffold does
   not install Biome (#441, filed from this work).

Every transcript in the README is from a run, and the paragraph under them says
which binary and which packages produced it.

## The same claim elsewhere

"The packages are not on npm" stood on seven pages under `docs/app/`. Each is
corrected where it stands, in one or two lines; `/reference/packages` already
carried the true version and the rest agree with it now.

## Claims checked, and how

| Claim | Method |
|---|---|
| 8 releases, 14 assets each, no Windows target | GitHub API `/releases` |
| 17 packages on npm, `alpha`=alpha.7, `latest` stale | a registry query per `packages/*` directory; matches `published-packages.txt` exactly |
| 11 more implemented and unpublished | `tools/release/pending-packages.txt` |
| The installer works, sha256 `ab8c17c6adc1` | ran `curl -fsSL https://setup.uniflowed.dev \| sh` into a scratch install root |
| create / install / build / test output | ran them; every transcript pasted is from that run |
| The scaffold pins the writing uf's version | `UNIFLOWED_VERSION = env!("CARGO_PKG_VERSION")`, plus the manifest it wrote |
| GraphQL is formatted | `crates/uf_fmt/src/graphql/{lexer,parser,printer,ast}.rs` and its fixtures |
| No bundler or dev-server crate | `ls crates/`; `uf_bundle`'s own module docs say it measures size |
| `dev.port` default | `crates/uf_config/src/lib.rs:147` — **5173**, not the 3000 the old README's example implied |
| Every relative link and both anchors resolve | a scripted existence check |

## Left for somebody to route

`docs/app/guide/scope/_uf.page.mdx` closes with "uf builds a static site today.
Everything that needs a server at run time works while `uf dev` is running and
disappears at `uf build`." That contradicts the table directly above it and
`/reference/cli`, both of which say `uf start`, `uf preview` and
`--adapter node` serve route handlers and render non-prerendered routes. Outside
this change's remit, so it is untouched.

## Exit codes

`uf fmt --check` 0 · `uf lint` 0 (0 errors, 16 warnings over 340 files, all
pre-existing) · `tools/docs/build.sh` 0 · `cargo test --workspace` 0 with
`UF_ALLOW_FIXTURE_SKIP=1` (2609 passed, 0 failed); without it the five `vite.rs`
cases that need a loopback socket fail, which is #420.

Closes #435

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/442"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791304303&installation_model_id=422833&pr_number=442&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F442&signature=87aa4812118ee3d9d0c3233be12029e62a7283301c2008ae8bf3a4568b9a33d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Note for whichever of this and #438 merges second.

The \"Where it stands\" list here says:

> **Server Components.** `\"use client\"` and `\"use server\"` are scanned, graphed and manifested; nothing that builds reads the result yet, so every page module still ships to the browser ([#252](https://github.com/ubugeeei-prod/uf/issues/252)).

#438 makes the second half false: a route no `\"use client\"` boundary reaches loses its page, layouts and loading boundaries from the client bundle. The first half stays true — nothing smaller than a route is split, and a Server Component above a boundary still ships whole, which is what keeps #252 open.

So the bullet needs to become something like: the analysis decides which *routes* reach the browser; a Server Component above a boundary still ships, and server actions are analysed and not callable.

PR #473, opened 2026-09-07
feat(rsc): a `"use server"` export the browser can call

ubugeeei:
## What was missing

`crates/uf_rsc` has scanned `"use server"`, keyed every action with a per-build
HMAC, decided which ones a client boundary can reach, published them in a
manifest and generated Flow types for them — since before anything could call
one. `crates/uf_rsc/src/types.rs:48` says the rest out loud:

> **A dispatcher.** There is no runtime that takes a name and calls an action,
> so this file declares no function.

This is the dispatcher, and the two halves either side of it.

## Which slice, and why not the other one

**Framing 1 from the brief: server actions callable end to end.** Framing 2 — a
Flight-shaped payload and a client that hydrates from it — is not a slice; it is
a second renderer, a wire format and a rewrite of `packages/router/client.js`,
and taking a piece of it would leave a half-built serialisation format in the
tree, which is the worst possible state for the thing whose whole risk is
deserialisation.

I did test #438's sentence rather than accept it.
`RscModule::requires_client_bundle`'s claim is true for a Server Component and
**was not** true for a `"use server"` module, which needs no payload at all —
the browser never renders it, only calls it.

## What this does

A `"use server"` module is replaced, **in the client graph only**, by one
`createServerReference` per callable export. What crosses the import is an id
and a `fetch`; the module's body, its imports and everything only they reached
stay on the server. The browser posts the id to the page's own URL, and an
endpoint between the middleware and the route handlers runs the function inside
the host's request.

One `load` hook decides that, so `uf dev` and `uf build` cannot disagree. One
`callAction` step, called between `runMiddleware` and `dispatch` in all four
hosts — `packages/vite/index.js`, `packages/vite/driver.js`,
`packages/server/fetch.js` and `packages/server/standalone.js` — so `uf dev`,
`uf preview`, `uf start`, a compiled binary and every deploy adapter answer it.

## Security is the feature

A `"use server"` export is a public HTTP endpoint the moment it exists.
`docs/security.md` gains nine rows and two sections; the boundary is:

> **A server action's arguments are plain JSON data and nothing else.** One JSON
> object `{"args": [...]}`, at most 1 MiB of valid UTF-8, holding at most 16
> values, nested at most 24 deep, with at most 10,000 values in total. Each is
> `null`, a boolean, a finite number, a string, an array of them, or a plain
> object whose keys are ordinary strings and are none of `__proto__`,
> `constructor` or `prototype`. The result travels back under the same grammar,
> plus `undefined`.

Nothing in a payload can name a function, a module, a class, a prototype, a
React element, an id or a reference, and nothing is revived into an object the
sender chose. Four things are outside it deliberately — a reference format,
class instances/`Map`/`Set`/`Date`, cycles, and `FormData` — each because
admitting it means admitting a tag that says which constructor to call.

Around that boundary:

- **Three independent CSRF guards.** `uf-action` is not a header a simple
  request may set, so a cross-origin caller needs a preflight and uf answers
  none; the content type must be `application/json`, which no `<form>` can
  produce; and `Origin` must be present and equal `Host`. `Host` only, never
  `X-Forwarded-*` — a forwarded header is a string the caller wrote.
- **No oracle.** One `404` with one body for a malformed id, an unknown id and
  an id that is not callable, over a table scanned whole with a constant-time
  comparison. The payload is decoded *before* the id is resolved, so a bad body
  cannot tell a real id from a guess.
- **Nothing about a failure goes back.** An action that throws is a `500` with a
  fixed body and a log line — the same in development, because `uf dev` and
  `uf build` have to agree about what this endpoint answers.
- **Bounds counted as the body arrives** rather than trusted from
  `Content-Length`, an iterative walk so the sender's hand is not on the stack
  depth, and `TextDecoder(…, { fatal: true })`.

And the honest limit, written where somebody will read it before deciding
otherwise: the call goes to the page's own URL so the middleware guarding that
path runs above it — but the URL is the caller's to choose, so **a server action
authorizes itself**.

## Flow checks both halves

Calling an action wrongly was already a `uf check` error and stays one. What a
declaration cannot say is whether those arguments can cross a wire at all —
`createUser(when: Date)` type-checks at every call site and arrives as `{}`. So
`uf prepare` writes two bounds into `server-actions.js`:

```js
export type ServerActionArgsFitTheWire = ActionArguments<ServerActionArgs<ServerActionName>>;
export type ServerActionResultsFitTheWire = ActionResult<ServerActionResult<ServerActionName>>;
```

Two lines hold a whole project against the grammar. A bound on `Parameters<F>`
rather than on `F`, because a function type puts its parameters contravariantly
and would ask whether `createUser` accepts *every* `ActionValue`.

## One rule in the graph, and its mirror

`propagate` already knew the server does not execute a `"use client"` module. It
now knows the client does not execute a `"use server"` one either, so an
action's imports are walked with the server colour. A database handle reached
only through an action is not in the client graph; `@uniflowed/server` imported
by an action is no longer reported as a leak, because it is no longer one; and a
`"use client"` module importing server-only code **directly** still is.

## Measured

`rsc-split-app`, one directive and one import different. On `main` an action
module that calls `cookies()` is a hard `rsc/server-only-import-in-client`
error, so the "before" column is the only shape that code could take.

| | dist raw | gzip | brotli | `/counter` chunk raw |
| --- | --- | --- | --- | --- |
| before | 213,742 | 68,236 | 59,257 | 1,389 |
| after | 216,545 | 69,533 | 60,361 | **4,194** |
| before, +22,437 B of server-only code behind the action | 231,260 | 71,600 | 61,746 | 18,899 |
| after, same | 216,553 | 69,528 | 60,418 | **4,194** |

The reference runtime is a **fixed** +2,803 raw / +1,297 gzip, paid once. After
that the client bundle stops growing with the server code behind an action:
+17,518 raw on `main` against **+8** with the reference, and at that size the
branch is 14,707 raw / 2,072 gzip / 1,328 brotli smaller. Break-even is about
2.8 kB raw of emitted JavaScript behind actions.

The bytes are the smaller half. `tally.js`, `ledger.js` and `node:async_hooks`
leave the browser outright: on `main` the first two are in
`dist/assets/_uf.page-*.js` and Vite prints `Module "node:async_hooks" has been
externalized for browser compatibility`.

## A bug found on the way

**`--compile` and `--adapter` never received `UF_RSC_MANIFEST`.** Harmless while
the manifest only split the route table — the deploy build is server-only, and
the server's table is unsplit — and not harmless now: `virtual:uf/actions` would
have been generated from no manifest, which is an empty table, which is **every
server action answering `404` in the artefact a person actually deploys**. Both
link steps now take one `LinkContext` rather than eight positional arguments,
which is the shape that let one of them be forgotten.

## Before and after

`cargo test -p uf_rsc`, with `graph/build.rs` and `types.rs` on `main`:

```
a_use_server_module_a_client_module_imports_stays_on_the_server ... FAILED
  left: ClientOnly / right: ServerOnly
a_server_only_import_reached_only_through_an_action_is_not_a_client_leak ... FAILED
  left: [ServerOnlyImportInClientModule { module: "app/actions.js", specifier: "@uniflowed/server" }]
a_project_with_an_action_holds_every_one_against_the_wire ... FAILED
test result: FAILED. 198 passed; 3 failed
```

after: `201 passed; 0 failed`.

`uf build` on the fixture, on `main`: `error: 1 React Server Components contract
violation`. After: `✓ build succeeded`, `server actions 1`, `rsc diagnostics 0`.

`uf test#library server-actions` with `packages/` on `main`: `failed to load:
SyntaxError … does not provide an export named 'ACTION_HEADER'`. After:
`37 passed`.

## Tests

- **`tests/library/server-actions.test.js`** — 37 cases. The grammar driven
  directly, because it is the argument boundary; the endpoint driven with no
  server, no port and no build the way `route-handler.test.js` drives
  `createDispatcher`, including the assertion that four different bad ids
  produce **one** distinct answer; the host ordering through the real
  `createFetchHandler`, so "below the guard, above the handlers" is a test
  rather than a comment; and the `load` hook once per environment.
- **`tests/type-tests/server-actions.js`** — six marked misuses, each `// expect:`
  written from what the checker actually said.
- **`crates/uf_cli/tests/vite.rs`** — one test reads the emitted production
  bundle (the reference is there; the action's marker, the ledger's marker and
  `async_hooks` are not; the server bundle has all three); one asks all four
  deploy artefacts to call a real action and compares their answers, including
  `"visitor":"ada"` from `cookies()`.

## What this does not do

- **A Server Component above a boundary still ships whole.** Framing 2 of #252,
  not started here.
- **An inline `"use server"` closure is still not callable** — no export name,
  so no reference. It stays in the manifest and out of both tables.
- **`<form action={fn}>`, `useActionState`, `useFormStatus`** need `FormData`
  across this boundary: multipart parsing, and a *simple* cross-origin request,
  which removes two of the three CSRF guards at once. Worth having; not worth
  having by accident.
- **`dist/uf-rsc-manifest.json` publishes the module graph.** Pre-existing, and
  it discloses no endpoint the client bundle did not already carry, but the
  paths and export names beside the ids are disclosure with no reader. Recorded
  as a `todo` row in `docs/security.md`.

## Exit codes

`cargo test --workspace` 0 (85 suites) · `cargo clippy --workspace --all-targets
-- -D warnings` 0 · `cargo fmt --check` 0 · `uf test#library` 0 (1894 passed, 1
skipped) · `uf fmt --check` 0 · `uf lint` 0 errors · `uf run scripts:parse` 0 ·
`uf run install:banner` 0 · `tools/docs/build.sh` 0.

Part of #252.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/473"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791335259&installation_model_id=422833&pr_number=473&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F473&signature=5e8bd91de0cf82f68a10e03b38293fbadeb3c56d9293618478e9890f7b490fc7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Red on `Library`, and it is a real interaction rather than a flake — `main` is green and so was this branch before it merged `main` in.

```
✗ cache.test.js  the route cache > renders once for two requests, and says so in a header
    error: TypeError: app.callAction is not a function
       --> cache.test.js:155:12
19 failed
```

#469 landed the route cache while this branch was open. Its tests build a host double and drive `createFetchHandler` over it; this branch adds a `callAction` step to that handler between `runMiddleware` and `dispatch`, so every double written before it now lacks a method the handler calls.

That is this branch's to fix, not #469's: the ordering change is the feature here, and #469's doubles were complete against the handler as it stood. The same edit was already made to the doubles in `deploy`, `serve`, `standalone` and `request-lifecycle` — `cache.test.js` is a fifth that did not exist when those were updated.

Worth deciding rather than patching five files a sixth time: a host double that has to be edited every time the handler grows a step is a double that will be wrong again. `request-lifecycle.test.js` already takes the position that a fake there proves only that the host calls something, and gives its double the real dispatcher over an empty table. The same answer would fit here.

No other check is failing, and the seven review threads are separate.

PR #544, opened 2026-09-07
feat(cell, state): state and derived, and no way back to cell and computed

### LINKED ISSUE #521: state: `computed` is called `derived`, and `cell` is called `state`
Two renames, and they are the same argument. `computed` is what the value *was*; `derived` is what it *is* — a value that follows from others, whether or not anything has computed it yet. And `@uniflowed/cell` names a mechanism where `@uniflowed/state` names the thing an application has.

They are not free: both are published names, `@uniflowed/state` already exists, and `@uniflowed/cell` is a dependency of packages that are also published.

**Done**: a decision on whether `cell` merges into `state` or is renamed beside it; the export renamed with the old spelling kept as a deprecated alias for one release; the codemod, because a rename a user has to do by hand is a rename most users do not do; and `tools/release/published-packages.txt` and the dependency closure check kept green through it.

Part of #488.

ubugeeei:
`@uniflowed/cell` built its two kinds of node with `cell` and `computed`, and
each named the machine rather than the value. `cell` named the node the call
allocates, at the one place a reader is asking a different question — does this
hold a value, or follow from one? `computed` named something that had already
happened, for a value that is usually computed later and sometimes never.
`state` and `derived` are the two answers to that question, so the pair now says
what the split is: everything in the graph is one or the other, and `effect` is
the derived cell whose value nobody reads.

## Before and after

```diff
-import { cell, computed, read, subscribe, write } from "@uniflowed/cell";
+import { derived, read, state, subscribe, write } from "@uniflowed/cell";

-const count = cell(0);
-const doubled = computed(() => read(count) * 2);
+const count = state(0);
+const doubled = derived(() => read(count) * 2);

 subscribe(doubled, () => console.log(read(doubled)));
 write(count, 21); // 42
```

Everything else in the package is untouched: `effect`, `resource`, `refresh`,
`status`, `read`, `peek`, `write`, `update`, `subscribe`, `snapshot`, `batch`,
`untracked`, and the `Cell<T>` type all keep their names. That snippet is the
one added to the state guide, and it was run before it was published — as a
temporary test file, which passed and was deleted.

## `cell` does not merge into `state`, and does not move

#521 left that open. Reading the two packages answers it, so here is the answer
rather than the option:

- **`@uniflowed/state` could not absorb it.** Five of its exports — `read`,
  `write`, `subscribe`, `refresh`, `batch` — already mean something else there,
  resolved against a store rather than against a value the cell holds. Merging
  would have had to rename those, which is a larger rename than this one.
- **It would invert a dependency.** `@uniflowed/state` depends on
  `@uniflowed/react`. `@uniflowed/cell` depends on nothing, and
  `@uniflowed/loader` reads cells on a server. Merging puts React in the
  dependency closure of the reactive core.
- **The package name is not the problem the issue thought it was.** `Cell<T>` is
  what both constructors return, `@uniflowed/loader` hands cells across a package
  boundary, and importing from this module at all means stepping below the atoms
  to work with the graph directly — which is the moment "a node in a dependency
  graph" is the useful word. An application reaches for `@uniflowed/state`;
  `@uniflowed/cell` is the layer under it, and is named for what it is.

## Backwards compatibility: no aliases, and no codemod

#521 asked for the old spellings as deprecated aliases for one release, and for a
codemod, on the premise that these are published names. **They are not.**

```
$ npm view @uniflowed/cell version
npm error 404 Not Found - GET https://registry.npmjs.org/@uniflowed%2fcell
$ npm view @uniflowed/state version
npm error 404 Not Found - GET https://registry.npmjs.org/@uniflowed%2fstate
```

Both sit in `tools/release/pending-packages.txt`, not in
`published-packages.txt`. So `cell` and `computed` have never been importable by
anybody, and the population an alias would serve is empty by construction. An
alias would be a second name for one thing — the objection `@uniflowed/state`'s
own module doc raises against `atomWithLazy` — and an alias promised for one
release is an alias still there three releases later. A codemod would have an
empty input set, and there is no codemod facility in the repository to hang one
on; if `@uniflowed/cell` ships and *then* renames something, that is when one
earns its place.

**So: a clean break, made on purpose.** The first published version of this
package exports `state` and `derived` and nothing that used to be called
something else. The decision is written down where somebody will meet it — the
module doc in `packages/cell/index.js`, and a new *The layer below* section in
the state guide — rather than only in this description.

`tools/release/published-packages.txt` does not name `cell`, so nothing moves
there, and `tools/release/verify-npm.sh --closure-only` stays green.

## One change this PR does not carry, and needs

`uf_lib`'s registry records what every `@uniflowed/*` module exports, and
`the_registry_names_exactly_what_each_package_exports` compares that list against
the package's real entry point. `@uniflowed/cell`'s entry still reads `"cell"`
and `"computed"`, so **that Rust test is red until this two-word change lands**
in `crates/uf_lib/src/registry.rs`:

```diff
             &[
                 "batch",
-                "cell",
-                "computed",
+                "derived",
                 "effect",
                 "peek",
                 "read",
                 "refresh",
                 "resource",
                 "snapshot",
+                "state",
                 "status",
                 "subscribe",
```

This branch was scoped to `packages/`, `tests/library/` and `docs/`, and could
not make it. Nothing else under `crates/` reads the shipped package: the
`@uniflowed/cell` fixtures in `uf_check` and `uf_cli` are written inline, and
`crates/uf_lib/tests/flow/effect-variance.js` imports only the `Cell` type, which
has not changed. `flow_exports` is read by that test and by `uf dev`'s hover
panel — nothing in `uf check` or `uf lint` consults it — so the effect of the
stale entry is a red test and a hover panel listing two names that no longer
exist.

## Verification

Run from the worktree with the prebuilt `uf 0.0.0-alpha.12`.

| Command | Result |
| --- | --- |
| `npm ci --no-audit --no-fund` | ok |
| `uf lint` | **passes** — `files checked 377, errors 0, warnings 14`, exit 0. The 14 warnings are pre-existing and identical to the baseline. |
| `uf fmt --check` | **passes** — `every file is already formatted`, exit 0. |
| `uf test tests/library` | **passes** — `1951 passed, 0 failed, 1 skipped in 5.13s`, exit 0. Byte-identical to the baseline taken on the unmodified tree. |
| `uf check` | **fails, and failed identically before this branch existed.** |
| `tools/release/verify-npm.sh --closure-only` | passes — `every @uniflowed/* dependency is itself published`. |

### About `uf check`

This is a pre-existing failure, unrelated to the change, and it is not small:
**1883 errors on the unmodified tree**, from the prebuilt binary carrying a newer
vendored Flow than this tree is written against. The clearest tell is 65
occurrences of

```
error[unsupported-syntax]: The + variance sigil is deprecated. Use readonly instead.
```

on a repository that uses `+` everywhere, alongside stricter `WeakMap` bounds and
`recursive-definition` inference that the tree predates. It is not papered over
and it is not fixable from here — fixing it means rebuilding the binary from this
tree, which needs cargo.

What can be said instead is that the change adds nothing to it. Scoped to the
files this touches:

```
uf check packages/cell packages/state packages/loader tests/library
  baseline  1395 errors
  branch    1395 errors
```

and the per-file breakdown is identical line for line —
`packages/state/index.js` 1, `internal/composed.js` 17, `internal/store.js` 3,
`tests/library/cell.test.js` 13, `tests/library/state.test.js` 16, before and
after. `uf check packages/cell` on its own is green: `7 files checked, 0 errors,
✓ no problems`.

Closes #521.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/544"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791368429&installation_model_id=422833&pr_number=544&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F544&signature=73c9d2c30f91ab58989732830438811549c3d97b2739e532d62a0fc918cc5839"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Picked up the one edit this PR left out. `crates/uf_lib/src/registry.rs` now names `state` and `derived`, so `the_registry_names_exactly_what_each_package_exports` is green — pushed onto this branch rather than filed as a follow-up, since the test is red until it lands. `cargo test -p uf_lib --lib`: 9 passed.

On the `uf check` noise: that binary was built from a different worktree's `upstream/flow`, which is why the counts do not line up with this tree. Your before/after comparison is the right way to have pinned it, and I have not treated it as this PR's problem.

PR #545, opened 2026-09-07
feat(pm): `uf pm approve-builds`, and the sentence npm gets instead

### LINKED ISSUE #495: pm: `uf approve-builds`, and the three spellings of trusting an install script
A dependency with an install script is arbitrary code at install time, and each manager treats it differently: pnpm asks (`pnpm approve-builds`), bun refuses until told (`bun pm trust`), npm runs everything unless `--ignore-scripts`, yarn berry has `enableScripts`.

This one is a security surface rather than a convenience, so the interesting work is the uf-side model: which packages this project has approved, recorded where the project can review it, rather than in each manager's own store.

**Done**: `uf approve-builds` lists what is waiting and approves by name; the approved set lives in `uf.config.js` or beside it; npm's "runs everything" is stated rather than papered over. See `docs/security.md`.

Part of #488.

ubugeeei:
Closes #495. Stacked on #543; the base retargets as that merges.

A dependency with a `postinstall` script is arbitrary code, run on the machine of everyone who installs it, before anybody has read a line of it. uf passes `--ignore-scripts` to every manager, so none of them runs — which is safe, and which also means `esbuild` and `sharp` do not work until somebody says so.

## What it shows

```
  manager       pnpm
  approvals in  pnpm.onlyBuiltDependencies

  package  version  runs         approved
  esbuild  0.24.0   postinstall  no
  sharp    0.33.5   postinstall  no

› 2 packages would run code at install time and are not approved, so uf does not let them run
› uf pm approve-builds <name>... records the ones you have read
```

It names the *hooks* — `postinstall` alone is a different risk from all four — and not the script bodies. A minified `postinstall` is one line of four kilobytes, and printing it would let the package decide how much of your terminal it gets; the file is one `cat` away, and reading it there is the review this command is asking you to do.

## Where the approval lives

| manager | field |
| --- | --- |
| pnpm | `pnpm.onlyBuiltDependencies` |
| bun | `trustedDependencies` |
| yarn 2+ | `dependenciesMeta.<name>.built` |
| npm, yarn 1 | — |

The field the project's own manager already reads. Not a file of uf's own and not a mirror of anybody's config schema: uf records the decision, the manager enforces it, a project that stops using uf keeps a working allow-list, and one that already had a list is read rather than overridden.

## What the install then does

`--ignore-scripts` comes off only when the manager can enforce a list **and** the list has something in it. An empty `onlyBuiltDependencies` is pnpm's way of saying "none", and dropping the flag for it would turn that into "whatever this manager defaults to" — a silent reopening of the hole the flag closes. Both call sites now go through one `scripts_allowed`, because a rule about running somebody else's code should have one spelling.

## npm

```
! npm cannot approve one and not another: `--ignore-scripts` is all of them or none
› pm.allowLifecycleScripts in uf.config.js turns on every one of them, deliberately
```

Stated rather than papered over, which is what #495 asked for. `uf pm approve-builds sharp` on npm is an error with the same explanation — an approval that quietly meant "and everything else too" would be worse than no command.

## Why `uf pm`

Because it is not something anybody runs daily. `uf install` and `uf add` are the verbs; this is a setting the manager reads that uf has an opinion about, and the top level is for the verbs. `uf pm` is the namespace for the rest of that plumbing.

## Two smaller decisions, both because this is a security surface

- **A name not in the tree is refused**, with the names that are. A typo that silently does nothing leaves the reader believing they approved something, and the eventual "why doesn't `sharp` work" gets debugged somewhere else entirely.
- **Approving does not install.** A command that reached for the install the moment you made the decision would run the script you were still thinking about.

## Also

`uf_pm::manifests::set` is the general form of the range splice from #541, for fields that are arrays and nested objects rather than one string. It refuses to overwrite a `"pnpm"` that is already something other than an object.

`docs/security.md` gains a "Dependency install scripts" section with the whole argument, and the lifecycle-scripts row now points at it.

## Verified

uf_pm 184, uf_cli lib 298, cli 66, dependencies 14, install 5, every_command 13. `cargo fmt --check`, clippy and `uf fmt --check` clean. Exercised end to end outside the repo: the two tables above are real, `pnpm.onlyBuiltDependencies` was written into a manifest whose indentation survived, a typo'd name was refused with the real ones, and the same project detected as npm produced npm's sentence.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/545"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791368566&installation_model_id=422833&pr_number=545&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F545&signature=4f3d602bd46dd549ea00837b2b93c0855077365ce87a79c5fe256ef0c3a52755"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/uf_cli/src/cli.rs
+++ b/crates/uf_cli/src/cli.rs
@@ -578,6 +578,16 @@ pub(crate) enum Commands {
         #[arg(long)]
         dry_run: bool,
     },
+    /// The package manager underneath: what it is allowed to do, and what it
+    /// has been told.
+    ///
+    /// `uf install`, `uf add` and the rest are the commands people run. This is
+    /// the plumbing beside them — the settings a manager reads that uf has an
+    /// opinion about.
+    Pm {
+        #[command(subcommand)]
+        command: PmCommand,
+    },
     /// Open a dependency for editing, and write the patch when you are done.
     ///
     /// `uf patch left-pad` prints a directory holding a copy of the package;
@@ -804,6 +814,31 @@ pub(crate) enum ReleaseBump {
     Patch,
     Minor,
     Major,
+}
+
+#[derive(Debug, Subcommand)]
+pub(crate) enum PmCommand {
+    /// List the dependencies that would run code at install time, and approve
+    /// the ones you have read.
+    ///
+    /// uf passes `--ignore-scripts` to every manager by default, so no
+    /// dependency runs anything. Naming one here records it in the field your
+    /// package manager reads — `pnpm.onlyBuiltDependencies`,
+    /// `trustedDependencies`, `dependenciesMeta` — and the next install builds
+    /// exactly those.
+    ///
+    /// npm and Yarn 1 have no per-package control: `--ignore-scripts` is all of
+    /// them or none. uf says so rather than offering an approval that quietly
+    /// means "and everything else too".
+    #[command(name = "approve-builds")]
+    ApproveBuilds {
+        /// The packages to approve; none lists what is waiting.
+        #[arg(value_name = "NAME")]
+        names: Vec<String>,
+        /// Say what would be approved, and write nothing.
+        #[arg(long)]
+        dry_run: bool,
+    },
 }
 
 #[derive(Debug, Subcommand)]
--- a/crates/uf_cli/src/commands/completion.rs
+++ b/crates/uf_cli/src/commands/completion.rs
@@ -35,6 +35,7 @@ const COMMANDS: &[&str] = &[
     "add",
     "audit",
     "build",
+    "pm",
     "patch",
     "catalog",
     "check",
@@ -104,6 +105,7 @@ const EXPLAINABLE: &[&str] = &[
     "remove",
     "update",
     "patch",
+    "pm",
     "catalog",
     "why",
     "ls",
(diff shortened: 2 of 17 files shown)


ubugeeei:
Moved from `uf approve-builds` to **`uf pm approve-builds`**. It is not a command anybody runs daily — `uf install` and `uf add` are the verbs, and this is a setting the manager reads that uf has an opinion about — so the top level was the wrong place for it. `uf pm` is now the namespace for the rest of that plumbing.

Everything else is unchanged. Re-verified: uf_cli lib 298, cli 66, every_command 13, `cargo fmt --check` and `uf fmt --check` clean, and the command re-run end to end under its new spelling.

Issue #526, opened 2026-09-07
lint: accessibility and markup rules, in the linter rather than beside it


ubugeeei:
uf's linter reads the official Flow parser's tree, so it sees JSX exactly. What it does not have is the rules `eslint-plugin-jsx-a11y` exists for, or the markup rules that catch what the HTML parser would silently rewrite.

The two are one issue because they are the same pass over the same tree, and because the second is a hydration bug rather than a style opinion: `<p><div>` is repaired by the browser before React sees it, and the repair is a mismatch.

**Done**: the a11y rules whose answer is unambiguous — an `img` with no `alt`, a click handler on a non-interactive element with no role or key handler, a label with no control, an `aria-*` that is not a real attribute, a heading level that skips — and the nesting rules the HTML parser will rewrite. Built in, not a plugin, because a rule that needs installing is a rule most projects do not have. And uf's own packages and the scaffold are clean under them.

Related: axe-core at runtime is #511; this is the half that needs no browser.

Part of #488.

PR #554, opened 2026-09-07
feat(temporal, web, hooks): Temporal on every host, and the two values a render has to fix

### LINKED ISSUE #509: web: time, randomness and markup that do not differ between the server and the client — and time means Temporal
Three things cause most hydration mismatches, and none of them is the application's bug in any interesting sense.

## Time, and it is Temporal rather than `Date`

`new Date()` on the server and on the client are different instants, and a page that formats one of them is a mismatch. That is the hydration half.

The other half is that **`Date` is the wrong API to build the answer on**. It is mutable, it is millisecond-only, it has one time zone — the host's — and it cannot say "a date with no time" or "a time in Tokyo" at all. Temporal is the standard that fixes those, `docs/roadmap.md` already lists "Lite Temporal" under P1, and a `<Time>` component built on `Date` today is a component that has to be rewritten when it lands.

So the deliverable is: **Temporal first, and the hydration-safe component falls out of it.** `packages/hooks/timing.js` holds `new Date()` and `Date.now()` in `useNow` and friends today, and `packages/effect` reads `Date.now()` for its schedules; those are the call sites.

* A `<Time>` that renders an absolute, zone-explicit instant on the server and localises after hydration — so the markup matches and the reader still sees their own zone.
* A clock the application gets from uf rather than from the host, so a test can move it and a server render is reproducible.
* Formatting through `Intl`, with the locale a decision rather than the host's.

## Randomness

`Math.random()` and `crypto.randomUUID()` differ by construction. `useId` covers ids; a seeded stream, generated once on the server and sent, covers the rest.

## Markup

Whitespace and invalid nesting — `<p><div>` — that the browser silently repairs before React ever sees it. The linter can catch the second at build time, and that half is [#526](https://github.com/ubugeeei-prod/uf/issues/526).

## What "done" means

* `Temporal` available to an application, from uf, on every host — including the ones without it natively, which is what "Lite Temporal" has to mean.
* `Time` in `@uniflowed/web`, built on it, with a test that renders it in a prerender and asserts the markup.
* A seeded random that survives serialization.
* `packages/hooks/timing.js` and `packages/effect` reading the injected clock rather than `Date.now()`.

Part of #488.

ubugeeei:
Time, randomness and markup that do not differ between the server and the client — and time means Temporal.

`Date` is the wrong thing to build a date component on: it is mutable, it is millisecond-only, it has exactly one time zone (the host's, and the two hosts here are set to different ones, which is the hydration problem restated), and it cannot represent a date with no time at all. So this is Temporal first, and the hydration-safe component falls out of it.

## The API

```jsx
// _uf.layout.js — once, above everything that reads a clock.
import { RenderProvider } from "@uniflowed/hooks";

export component Layout({ children }) {
  return <RenderProvider>{children}</RenderProvider>;
}
```

`RenderProvider` fixes three values once — the instant, the zone and a seed — writes them into the markup as an inert `<script type="application/json">`, and reads them back on the client before its first render. Neither side has to be told which one it is.

```jsx
import { Time } from "@uniflowed/web";
import { useRandom, useRenderedAt, useShuffled } from "@uniflowed/hooks";
import { Temporal } from "@uniflowed/core/temporal";

component Article({ postedAt, related }) {
  // Constant on both sides. Not "now", and it never becomes "now".
  const renderedAt = useRenderedAt();              // Temporal.Instant
  const age = renderedAt.since(Temporal.Instant.from(postedAt));

  return (
    <article>
      {/* iso, date and zoned are the same string everywhere.        */}
      {/* local and relative render the deterministic form first and */}
      {/* become the reader's after hydration.                       */}
      <Time value={postedAt} format="local" />
      <p>{age.total({ unit: "hour" }).toFixed(0)} hours old</p>

      {/* Same order on the server and in the browser. */}
      <ul>
        {useShuffled(related, "related").map((post) => (
          <li key={post.id}>{post.title}</li>
        ))}
      </ul>
    </article>
  );
}
```

The server writes `<time datetime="2026-09-04T06:00:00Z">2026-09-04 15:00 +09:00</time>` — an absolute instant, zone-explicit, readable before any JavaScript runs — and the browser replaces the text with the reader's own locale and zone after hydration. `useRandom(label)` returns a stream; `label` is what makes it independent of every other one.

Testing a page is installing a clock:

```js
import { fixedClock, setClock } from "@uniflowed/core/clock";

const restore = setClock(fixedClock(Date.UTC(2026, 8, 4, 6, 0, 0), "Asia/Tokyo"));
// Temporal.Now, useNow, useTimeAgo, a retry timetable — all of them read this.
restore(); // puts back what it found, not "the system clock"
```

## What Temporal support requires of the runtime, and what happens without it

**Nothing.** `@uniflowed/core/temporal` uses `globalThis.Temporal` where the host has it — Node 24.20, which this repository's tests run on, does not — and where it does not, it implements `Instant`, `ZonedDateTime`, `PlainDate`, `PlainTime` and `Duration` over `Intl.DateTimeFormat`, which is where a JavaScript host keeps its zone data whether or not it has shipped Temporal. Those are the five types `@uniflowed/temporal` has named since it was a declaration.

The documented subset is the contract: everything in it is spec-shaped, so code written against it runs unchanged on both. What is deliberately absent, so nobody discovers it by having it throw:

* **Non-ISO calendars.** Japanese, Hebrew and Islamic *arithmetic* is the CLDR calendar tables. `@uniflowed/temporal`'s `withCalendar` raises `NativeRuntimeRequiredError`, naming itself, which is what every uf declaration does for a binding that needs the binary. Formatting in another calendar — `toLocaleString(locale, { calendar: "japanese" })` — goes through `Intl` and works today, and is what most callers asking for this actually want.
* **Nanoseconds.** The clock seam is milliseconds because that is what a host can promise, so `epochNanoseconds` is absent rather than offered as a number with three invented zeros.
* **`round`, `with`, `PlainDateTime`, `PlainYearMonth`, `PlainMonthDay`.** Nothing uf ships needs them, and a half-implemented calendar type is worse than an absent one: absent is a name error at build time.

Two smaller consequences, both named in the source: on a host without Temporal, `x instanceof Temporal.Instant` is false (the exported surface is written out one entry at a time so a method added to `LiteInstant` cannot silently widen what uf promises), and a host with no `Intl` at all can still render UTC, because `partsIn` answers that zone from `Date` without asking for zone data it would not have.

`Temporal.Now` is uf's in **both** cases. Native `Temporal.Now` reads the system clock, and a `Temporal.Now.instant()` inside a render is the bug this whole area exists to fix — deferring for `Now` would have made a page's correctness depend on whether the browser had shipped Temporal yet.

## Where the code lives

The implementation is `@uniflowed/core/temporal`, `@uniflowed/core/clock` and `@uniflowed/core/random`; `@uniflowed/temporal` re-exports it. That is a constraint rather than a preference: `@uniflowed/hooks` reads the same clock and is on npm, `@uniflowed/temporal` is not, and `tools/ci/publishable.sh` refuses a published package that depends on an unpublished one — `npm install @uniflowed/hooks` would answer `ETARGET`. The opaque types in `packages/temporal/index.js` already named `@uniflowed/core/temporal` as their module, so this is the layering that was written down rather than a new one.

## The rest of the issue

**The injected clock.** `packages/hooks/timing.js` (`useNow`, `useTimeAgo`, `useThrottledCallback`) and `packages/effect` (`retry`, `repeat`) read `@uniflowed/core/clock` instead of `Date.now()`, and `retry`/`repeat` draw their jitter from `@uniflowed/core/random` instead of `Math.random()`. `schedule.js` has always taken both as arguments; this is the other end of that decision, and until now the other end was the host.

**Randomness.** sfc32 over a 128-bit string hash — four words of integer arithmetic, because that is the only arithmetic JavaScript specifies exactly and both renders must produce the same numbers byte for byte. `fork(label)` is the part worth reading: a single shared stream makes every consumer depend on the draw order, that order is the render order, and the render order is *not* stable once a tree suspends. The bug that produces is correct in development, correct in a test, and wrong on a slow connection.

**Markup.** The zone half is covered — the deterministic text is a function of the instant and a zone name rather than of the machine. Invalid nesting (`<p><div>`) is #526 and is not touched here.

## Verification

Run from the worktree, against `target/debug/uf` at `0.0.0-alpha.12`:

| Command | Result |
| --- | --- |
| `npm ci --no-audit --no-fund` | clean |
| `uf lint` | **0 errors, 14 warnings** — the same fourteen as `main` |
| `uf fmt --check` | **every file is already formatted** |
| `uf test tests/library` | **2002 passed, 0 failed, 1 skipped** (up from 1951 on `main`) |
| `uf check` | **fails with 1883 errors** |

`uf check` failing is pre-existing and is stated rather than papered over: `main` fails with **1883 errors and 14 warnings**, this branch fails with **1883 errors and 14 warnings**, and the per-file table is identical — every file this branch touches is at its `main` count, and the four new modules and the new test file contribute none. `uf check` is not in `ci.dependsOn` in `uf.config.js`; the checker does not resolve types across modules yet, which is what most of those 1883 are. Reaching that number took two rounds of real fixes — the five Temporal value types are Flow `interface`s rather than object types, because a class instance is not a subtype of an object type, and `DateTimeFormatOptions` is `readonly` throughout, because an optional writable property is invariant.

Also run: `tools/ci/publishable.sh` (26 implemented packages, 15 published, 11 waiting on `npm trust`), `tools/release/verify-npm.sh --closure-only` (closed), and `tools/ci/lockfile-in-sync.sh` (matches all 50 manifests, after regenerating `package-lock.json` — three manifests gained a dependency, and that check is what refuses the alternative).

New tests: `tests/library/temporal.test.js` (32 cases — the clock seam, the five types, `Now`, the seeded stream; the zone cases name Tokyo for a date that differs from UTC's and Kathmandu for the `+05:45` an offset computed in whole hours gets wrong), a prerender block in `tests/library/web.test.js` that renders `Time` through `react-dom/server` and asserts the bytes, a `the render anchor` block in `tests/library/hooks.test.js`, a `the render anchor, prerendered` block in `tests/library/hooks-ssr.test.js` — the only process with no document, which is the side of the handoff that goes first — and two cases in `tests/library/effect.test.js` for the clock and the jitter.

## Worth a follow-up

* `@uniflowed/temporal` is in neither release manifest, so `@uniflowed/core/temporal` is the specifier that installs. Adding it to `tools/release/pending-packages.txt` needs a file outside this change's scope.
* `RenderProvider` has to be rendered by the application. The router already writes a `__uf_data` script and reads it back in `hydrate()`, and that is where this could be automatic.
* `packages/web/time.js` still uses `Intl.RelativeTimeFormat` directly for `relative`; `useTimeAgo` declares its own shape for it because Flow's vendored `intl.js` does not.

Closes #509.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/554"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791371884&installation_model_id=422833&pr_number=554&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F554&signature=8fc5938d5627f4ff2df2e0c6d01f593eca95774223500e990e73075acb7d1c84"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Applied the `crates/uf_lib/src/registry.rs` edits — three packages needed names, not one: the render envelope in `@uniflowed/hooks` (`RENDER_ID`, `RenderProvider`, `useRandom`, `useRenderEnvelope`, `useRenderTimeZone`, `useRenderedAt`, `useShuffled`), `asInstant` in `@uniflowed/web`, and `isLite`/`withCalendar` in `@uniflowed/temporal`. `cargo test -p uf_lib --lib` 9 passed. Auto-merge armed.

The `package-lock.json` regeneration is right and needed no flag — `lockfile-in-sync.sh` is a CI gate and three manifests gained a dependency.

Your follow-ups 1 and 2 are real: `@uniflowed/temporal` needs to reach `pending-packages.txt`, and `RenderProvider` belongs in the router's `hydrate()` beside the `__uf_data` script rather than being something every project wires by hand. I will file both.

PR #555, opened 2026-09-07
feat(web): the five numbers a page is judged on, read from the browser

### LINKED ISSUE #510: web: web vitals, measured and reported
LCP, INP, CLS and TTFB are what a real application is judged on, and uf ships no way to read them.

**Done**: an opt-in collector in `@uniflowed/web` that reports the four through `PerformanceObserver`, an endpoint contract for sending them somewhere (rather than an integration with an analytics company), and `uf dev` printing them for the page being developed — which is the half a third-party service cannot give you, because it is about the change you just made.

Part of #488.

ubugeeei:
LCP, CLS, INP, TTFB and FCP, read straight out of the browser's own
`PerformanceObserver` in `@uniflowed/web`, plus the endpoint contract for
sending them somewhere that is not an analytics company.

## The API

```js
import { collectVitals, useVitals, vitalsBeacon } from "@uniflowed/web";

// Imperative: returns a function that stops collecting.
collectVitals({
  report: (vital) => {
    console.log(vital.name, vital.value, vital.rating);
    // LCP 1840 good
  },
});

// From a root layout. Same thing, tied to a mount.
export component Layout(children: React.Node) {
  useVitals((vital) => console.log(vital));
  return <>{children}</>;
}

// Post them to a path on your own origin.
collectVitals({ report: vitalsBeacon("/api/vitals") });
```

```js
// app/api/vitals/_uf.route.js
export async function POST(request: Request): Promise<Response> {
  const report = await request.json();
  // { url: "https://example.com/pricing", vitals: [ … ] }
  return new Response(null, { status: 204 });
}
```

Each metric arrives once, when it is final:

```js
{ name: "LCP", value: 1840, rating: "good", navigationType: "navigate" }
```

New exports: `collectVitals`, `useVitals`, `vitalsBeacon`, `VITALS_ENDPOINT`,
and the types `Vital`, `VitalName`, `VitalsReporter`, `VitalsReport`, `Rating`,
`NavigationType`, `CollectOptions`. New subpath `@uniflowed/web/vitals`.

## What leaves the machine by default

**Nothing.**

- Importing `@uniflowed/web` or `@uniflowed/web/vitals` does nothing at all —
  no top-level side effect, no listener, no connection, nothing for a bundler
  to keep.
- `collectVitals` calls the function you gave it. That is the entire effect.
- `vitalsBeacon` has to be constructed by the project, and takes a **path**
  rather than a URL, so it cannot be aimed at another origin by accident.
- The default path is `/__uf/vitals`, same-origin, and nothing serves it unless
  the project chooses to. There is no uf endpoint, no third party, and no
  analytics integration to switch off.

## Design notes

**Direct reads, not a vendored `web-vitals`.** The corrections that turn a raw
entry stream into a number anybody quotes are implemented rather than
approximated: CLS is the worst five-second session window (1s gap, 5s span)
rather than a sum, INP groups events by `interactionId` and steps one place
down the ten longest for every fifty interactions, and `activationStart` is
subtracted at the moment a metric is *reported* rather than when its entry
arrived — because that field is zero until a prerendered page is activated.

**Documented differences from `web-vitals`** (`docs/app/guide/vitals`): one set
of metrics per page load rather than re-arming on a bfcache restore; no
attribution build; no `delta`; prerendering corrected rather than deferred; and
where `performance.interactionCount` is missing, the observed interaction count
is the denominator, which biases INP towards the slowest interaction rather
than away from it.

**Absent, never zero.** A metric whose entry type the browser does not serve is
not reported. Zero is the best possible score for three of these five, so
reporting it for "no idea" makes a site's worst pages look like its best ones,
and it looks like good news. CLS *is* reported as zero when `layout-shift` was
served and nothing moved — that is a measurement. INP is not reported when
nobody interacted — there is nothing that could have been slow.

**The gate is `supportedEntryTypes`, never `typeof PerformanceObserver`.** Node
has had a `PerformanceObserver` for years and serves none of these entry types,
so the obvious guard passes during a server render and then measures nothing.
Asking the constructor which types it serves answers "is there a browser", "is
it too old" and "is this a test host that borrowed Node's constructor" in one
expression — which is also why the "reports nothing" tests need no stubbing at
all: `uf test` runs on exactly that host.

**SSR-safe.** All three entry points return without doing anything where there
is no document, and none of them throws. uf prerenders every static route, so
each runs with no browser on every build.

**More than one report per page load.** Metrics become final at different
moments and the beacon coalesces across a microtask instead of holding
everything until the page is hidden. Batching to one request would have made
the whole thing depend on which page-hide listener the browser called first —
the collector's, which produces the last three metrics, or the beacon's, which
sends them — and that ordering is not something either side can state.

## The `crates/` diff I could not apply

`crates/uf_lib/src/registry.rs` mirrors every package's entry-point export list
and `the_registry_names_exactly_what_each_package_exports` compares the two.
This PR adds four values to `packages/web/index.js`, so that test will fail
until this is applied. I could not edit `crates/`.

```diff
--- a/crates/uf_lib/src/registry.rs
+++ b/crates/uf_lib/src/registry.rs
@@ -434,10 +434,13 @@
         NativeModule::new(
             "@uniflowed/web",
             NativeModuleKind::Framework,
             Stability::Experimental,
             &[
                 "Announcer",
                 "Font",
                 "Image",
                 "Layout",
                 "Page",
                 "Picture",
                 "SkipLink",
                 "Time",
+                "VITALS_ENDPOINT",
+                "collectVitals",
                 "relative",
                 "useCookie",
                 "useHead",
+                "useVitals",
+                "vitalsBeacon",
             ],
         ),
```

That is the full sorted list, for reference:

```
Announcer, Font, Image, Layout, Page, Picture, SkipLink, Time,
VITALS_ENDPOINT, collectVitals, relative, useCookie, useHead, useVitals,
vitalsBeacon
```

Types are deliberately not in it, per that file's own rule.

## Not in this PR

**`uf dev` does not answer `/__uf/vitals`.** That is the third bullet of the
issue and it is the half a third-party service cannot give you, so it is worth
being explicit rather than quiet: the browser half of the contract is finished
and the receiver is not. It belongs to the dev server, which is `crates/`, and
the path is deliberately in the namespace the dev server already owns —
`/__uf/`, beside `@uniflowed/hmr`'s update stream, and unreachable by any route
a project writes because an `app/` directory beginning with `_` is not a route.
`docs/app/guide/vitals` says so and gives `console.log` as the interim answer.

The receiver is now a small, fully specified job: accept `POST /__uf/vitals`
with a `{ url, vitals: [{ name, value, rating, navigationType }] }` body and
print each metric with its rating beside the reload that produced it.

Worth separate issues:

- The dev-server receiver and the terminal print, above.
- **Attribution.** `web-vitals` has a second build that names the LCP element,
  the shifts that made up the CLS and the target of the slowest interaction.
  For a number you are reading about your own page, that is the actionable
  half, and it is missing. It needs a way to put a DOM node in a JSON payload,
  which is a design decision rather than an omission.
- **bfcache restores.** A restore is a second page view inside one document and
  neither `Vital` nor `VitalsReport` can say which view a number belongs to.
  Adding that is a change to the endpoint contract, not to the collector.

## Verification

Run from the worktree, with the pre-built binary.

```
$ uf fmt --check

$ uf lint
  files checked  380
  errors         0
  warnings       14
! 14 warnings

$ uf test tests/library
✓ 1993 passed, 0 failed, 1 skipped in 35.84s

$ uf test tests/library/web-vitals.test.js
✓ 42 passed, 0 failed in 1.46s
```

Baselines taken on the untouched tree before any change:

| | before | after |
| --- | --- | --- |
| `uf test tests/library` | 1951 passed, 0 failed, 1 skipped | 1993 passed, 0 failed, 1 skipped |
| `uf lint` | 0 errors, 14 warnings | 0 errors, 14 warnings |
| `uf fmt --check` | clean | clean |
| `uf check` | 1883 errors | **1883 errors** |

`uf check` is unchanged, and `uf check 2>&1 | grep -i vitals` matches nothing —
neither of the new files contributes an error. (One intermediate run reported
1886; those three were mine, all fixed: a `(expr: Type)` cast that should be
`expr as Type`, and an exact-object prop mismatch in a test component.)

The 42 new tests are in `tests/library/web-vitals.test.js`. The ones that carry
the most weight:

- `a browser without the entry type reports nothing rather than zero` — run
  against Node's real `PerformanceObserver`, with nothing stubbed.
- `reports a zero cls when the browser measured no shift at all` — the other
  side of that rule.
- `reports the worst run of shifts rather than their total` and `ends a run
  after five seconds even with no gap in it` — the two halves of the session
  window.
- `steps down from the slowest once a page has had fifty interactions` — the
  INP percentile, which a maximum would pass and a sum would not.
- `reads the activation at the moment it reports, not when the entry arrived` —
  the prerender correction, which is wrong in a way that looks right if applied
  when the entry lands.
- `reports nothing at all once it has been stopped` — the `StrictMode`
  double-mount guard.
- `falls back to a keepalive post when the beacon queue refuses the report`.

`docs/app/guide/vitals` was built with `uf build` in `docs/` — the page renders
and is in the sitemap.

Closes #510.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/555"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791371906&installation_model_id=422833&pr_number=555&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F555&signature=c232a4ae211d8bc85b1b504bea561e86b62bfbbbc82e5800a504a9e145927532"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Applied the `crates/uf_lib/src/registry.rs` edit you flagged — four names in ASCII-sorted position, `cargo test -p uf_lib --lib` 9 passed. Auto-merge is armed.

Filed your first follow-up as part of the record: the dev-server receiver for `/__uf/vitals` is the remaining half and it is a `crates/` job, so it was correctly out of your scope.

Issue #561, opened 2026-09-07
lib, ui: nothing compares `uf_lib`'s parts lists against the components the package ships


ubugeeei:
`crates/uf_lib/src/registry.rs` mirrors every package's *export* list and `the_registry_names_exactly_what_each_package_exports` holds it to the real entry point. `crates/uf_lib/src/ui.rs` mirrors every component's *parts* list and nothing holds it to anything.

So `Combobox` grew `Group` and `GroupLabel` in [#558](https://github.com/ubugeeei-prod/uf/pull/558) and the Rust list still said seven parts, with every test passing. It was found by reading, which is not a check.

**Done**: a test of the same shape as the registry's — read the namespace object each component exports, compare it against the parts named here, and fail naming both directions. `Command` is the interesting case: it is listed and deliberately not implemented, so the test has to know the difference between "absent because nobody wrote it" and "absent on purpose", which means that fact belongs in the data rather than in a comment.

Found while merging #558.

PR #563, opened 2026-09-07
feat(server): QUERY, event streams, an upgrade a handler can accept, and a queue

### LINKED ISSUE #512: server: the HTTP QUERY method
QUERY is a safe, idempotent method with a request body — the thing `GET` with a long query string is a workaround for. Route handlers accept the fixed set of verbs and QUERY is not in it.

**Done**: `QUERY` is a handler a route can export; the router matches it; the client helper sends it; and what happens on a runtime or a proxy that refuses the method is stated rather than discovered — which is most of the work, because that is the reason it is not already everywhere.

Part of #488.

### LINKED ISSUE #529: server: SSE, WebSockets, and a queue
Three things a real application needs from its server that uf has no answer for.

* **Server-sent events.** A route handler that returns a stream, and a client hook that reconnects. The smallest of the three and the one most applications actually want.
* **WebSockets.** An upgrade a route handler can accept, which the adapters have to agree about: `node` can, `edge` can with a different API, and `serverless` mostly cannot. That disagreement is the design work.
* **A queue.** Work that outlives a request. `after()` exists and runs in-process — #389 says it runs *before* the response is sent, which is its own bug — and a queue is the durable version: a job, a handler, a retry policy, and a backend contract with one implementation.

Each needs the capability contract `docs/security.md` and the adapter matrix already ask for: a deployment target that cannot hold a socket open must refuse the configuration rather than accept it and drop connections.

Part of #488, and adjacent to #391 (six adapters) and #389.

ubugeeei:
Two issues in one pull request, because they are one seam. `QUERY`, event
streams, WebSockets and a queue are all a route handler asking the host for
something the host may not have, and splitting them would have meant designing
that seam twice.

### `QUERY` — a `GET` with a body

A handler exports it like any other verb:

```js
// app/api/search/_uf.route.js
// @flow
export async function QUERY(request: Request): Promise<Response> {
  return Response.json(await search(await request.json()));
}
```

and the client helper sends one:

```js
const hits = await api.request("/search", { method: "QUERY", body: { filters } });
```

It is retried like the other safe methods, and it appears in the `Allow` header
of a 405 so a client can tell it is offered.

**What refuses it is most of the work.** Nothing in uf does. A proxy, CDN or
WAF with a list of verbs answers `405` or `501` itself, so the request never
arrives and the failure looks exactly like a route that does not exist — from a
server that never got the chance to have one. `@uniflowed/fetch` says that in
the error rather than passing the status through. A cache in front of the
application must not store a `QUERY`; uf's own route cache stays `GET`-only.
`<form>`, `XMLHttpRequest` and `EventSource` cannot send it and `fetch` can.

There is no method-override header and there will not be:
`X-HTTP-Method-Override` on a `POST` is an inbound header steering dispatch,
which is the shape of CVE-2025-29927 and is forbidden by `docs/security.md`.

The verb list was also closed in name only — `pick` looked the method up on the
module, so a module exporting `PURGE` answered a `PURGE`. It is closed now.

### Server-sent events

```js
// app/api/progress/_uf.route.js
// @flow
import { eventStream } from "@uniflowed/server/events";

export function GET(): Response {
  return eventStream((sink) => {
    const timer = setInterval(() => sink.send({ event: "tick", data: percent() }), 1000);
    return () => clearInterval(timer);
  });
}
```

It is a `Response`, so it needs nothing new from the dispatcher. `eventStream`
writes the format, sets the headers that stop a proxy holding the whole stream
until it ends, sends a heartbeat so an idle connection is not closed by
something in the middle, runs the source's cleanup when the client hangs up,
and disconnects a reader that falls more than twice the high-water mark behind
rather than accumulating it.

The client half:

```js
import { useEventSource } from "@uniflowed/hooks/events";

const stream = useEventSource("/api/progress", { events: ["tick"] });
// stream.status: "idle" | "connecting" | "open" | "closed"
// stream.last:   { name, data, id } | null
```

`EventSource` reconnects on its own and needs no help — except in the one case
where it gives up permanently: a response that is not a `200` with
`content-type: text/event-stream`, which is what a deployment restarting looks
like. The hook reopens that on a doubling delay, stays out of the way while the
browser is retrying by itself, and reports the connection as state. It is
`"idle"` on the server and on the first client render, and opens nothing during
a render.

### WebSockets

```js
// app/api/room/[id]/_uf.route.js
// @flow
import { upgradeWebSocket } from "@uniflowed/server/socket";

export function GET(request: Request, context: HandlerContext): Response {
  const { response, socket } = upgradeWebSocket(request);
  socket.addEventListener("message", (event) => socket.send(String(event.data)));
  return response;
}
```

**uf defines the upgrade and the deployment supplies it.** Node has no
server-side `WebSocket` at all; Deno answers `{ response, socket }`; Bun answers
a boolean and puts the handlers on the server; Cloudflare wants a
`WebSocketPair` and a `Response` carrying `webSocket`; a Lambda cannot. Four
spellings and one refusal, and no naming makes them one API — so uf implements
none of them and owns the shape they reduce to, `(request) => { response,
socket }`, passed as `websocket` where the host is built:

```js
createFetchHandler({ app, document, capabilities: nodeCapabilities({ websocket }) });
```

A built-in wrapping `ws` would make uf the thing that decides which WebSocket
library a project has, which is what `docs/red-lines.md` rule 3 exists to
prevent.

### A queue

```js
// @flow
import { defineJob, enqueue } from "@uniflowed/server/queue";

export const sendWelcome = defineJob({
  name: "send-welcome",
  retry: { attempts: 5 },
  run: async (payload) => mail(payload.to),
});

export async function POST(request: Request): Promise<Response> {
  const user = await create(await request.json());
  await enqueue(sendWelcome, { to: user.email });
  return Response.json(user, { status: 201 });
}
```

## What a deployment has to provide for the queue

**uf provides:** the job and its name, so a worker in another process can find
the function again; the payload boundary, serialised as JSON where it is
enqueued rather than where it is stored, so the in-process backend behaves
exactly like a durable one; the retry policy and its backoff; `createRunner`,
which is the function a consumer calls per record and where the retry decision
is made, so every backend retries the same way; the five-field record shape, so
an implementation over SQS, a Redis list or a Postgres table is a `push` and a
loop rather than an adapter; and one backend, `memoryQueue`, that runs in this
process.

**The deployment provides two things no configuration flag can:**

1. **Durable storage.** `memoryQueue` is an array. A restart loses everything
   in it, two instances behind a load balancer each have their own, and nothing
   is acknowledged. That is a legitimate small deployment — one long-lived
   process, work that can be lost — and it says `durable: false` on the value
   so an adapter can refuse it rather than a person having to have read a
   paragraph.
2. **A consumer: a process that exists when no request is being answered.** A
   container or a long-lived Node process has one. A serverless function does
   not — it is frozen between invocations. A worker isolate does not either;
   the platform may tear it down the moment the response is written, which is
   the same reason `after()` there goes through `ctx.waitUntil`. On both, the
   storage has to be elsewhere *and* the drain has to be something else: a
   scheduled invocation, a queue-triggered function, a container beside the
   site.

Delivery is at least once — a record handed to a runner whose process then dies
is a record the backend still holds — so a job that charges a card needs its own
idempotency key. That is stated in the module rather than left to be discovered,
because it cannot be fixed later by a library.

## What each target can do, and what it refuses

| Target | Streams | Outlives the response | Event streams | WebSockets | In-process queue |
| --- | --- | --- | --- | --- | --- |
| `uf dev`, `uf preview`, `uf start` | yes | yes | yes | with an upgrader | yes |
| `node`, `container`, `--compile` | yes | yes | yes | with an upgrader | yes |
| `edge` | yes | no | yes | with an upgrader | **refused** |
| `serverless` | no | no | **refused** | **refused** | **refused** |

Refused at wiring time where it can be — `lambdaCapabilities({ websocket })`
throws, so a deployment that would accept handshakes and drop every one fails to
start — and at the point the handler asks where it cannot, with the target
named, rather than as a connection nobody serves.

## Verification

Run from the worktree, with the `uf` binary from
`.claude/worktrees/vite-typescript-nextjs-toolchain-c3f5e2/target/debug/uf`:

| Command | Result |
| --- | --- |
| `uf test tests/library` | **2000 passed, 0 failed, 1 skipped**, 51 files, 19.8s |
| `uf lint` | **0 errors, 14 warnings** — the same fourteen `main` reports |
| `uf fmt --check` | **every file is already formatted** |
| `uf check` | **fails, and fails identically on `main`** — see below |

**`uf check` fails on `main` and this change does not move it.** Measured cold
(`rm -rf .uf/cache/check`) and back to back on the same machine: `main` reports
**1883 errors**, this branch reports **1883 errors**. The first version of this
branch added 22 — a deprecated `+` variance sigil, an unannotated empty array, a
structural stream-controller type a class instance cannot satisfy, and a handful
of missing annotations in the new suite — and every one of them is fixed in the
commit. `uf check` is not in `uf run ci`, which is why the 1883 stand.

Two things about `uf check` are worth reporting rather than papering over, and
neither is caused by this change:

- **Its results are not stable across runs.** The same tree reported 3439 and
  1883 in two sessions an hour apart, and a warm cache and a cold one disagreed
  by 1600 on one tree. The per-file numbers move with them. Only a back-to-back
  cold pair is comparable, which is what the paragraph above is.
- **It intermittently exceeds its own inference budget** on
  `tests/library/ui.test.js`, a file this branch does not touch, and stops the
  run: `type inference for tests/library/ui.test.js exceeded its 36240ms
  budget`. It happened twice on a loaded machine and not on an idle one.

44 new test cases, in `tests/library/transports.test.js` (the three server
modules and the capability refusals), `route-handler.test.js` (QUERY dispatch,
`Allow`, and the verb that is no longer answered), `fetch.test.js` (sending,
retrying and the 405 message), `hooks.test.js` (the hook's four cases, including
the reconnect the browser will not make) and `hooks-ssr.test.js` (it opens
nothing during a prerender).

Closes #512.
Closes #529.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/563"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791372177&installation_model_id=422833&pr_number=563&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F563&signature=02e0cbf8c5594b7f37f86de5eb9b204be0b4b24805701af29aace82404f88153"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Applied the `registry.rs` edit — `useEventSource` in sorted position, `cargo test -p uf_lib --lib` 9 passed. Auto-merge armed.

One correction to your point 3: `registry.rs` **is** cross-checked — `the_registry_names_exactly_what_each_package_exports` compares it against every package's real entry point, which is exactly how your `useEventSource` drift surfaced. The list that is *not* checked is `crates/uf_lib/src/ui.rs`'s component parts, which I filed as #561 an hour ago after the same thing happened there.

Your points 1 and 2 about `uf check` are serious and I am filing both — a checker whose answer depends on cache state is a checker nobody can act on, and a wall-clock inference budget that fails under load is a CI flake waiting to happen. Thank you for measuring cold-vs-cold rather than reporting the raw numbers.

PR #567, opened 2026-09-07
feat(i18n): the arguments a message takes, in the type

### LINKED ISSUE #530: std: type-safe i18n, built on MessageFormat 2
MF2 is the current standard — a formal syntax for plurals, selection and markup — and it lands with typed placeholders, which is what makes it worth building on rather than around: a message with a `{$count}` has a type, and a call that omits it should not compile.

**Done**: a Flow-typed catalogue where a message key is a union rather than a string, its arguments are inferred from the message's own placeholders, and a missing or misspelled key is a type error; MF2 parsed at build time so the runtime carries a compiled form; locale negotiation, and a lazily loaded catalogue per locale so a page ships one.

The reason to do it here rather than reach for a library: every existing one types the key and gives up on the arguments, because in TypeScript that needs the message parsed at type level. uf parses the message at build time already.

Part of #488.

ubugeeei:
`@uniflowed/i18n`: MessageFormat 2 with the arguments a message takes in the type.

Every i18n library types the key and gives up on the arguments, so a message that gained a `{$count}` in the source locale keeps compiling at every call site and renders the placeholder to a user. This closes that.

## The API

```js
// @flow
import { defineCatalogue, message, number, string } from "@uniflowed/i18n";

const messages = {
  greeting: message("Hello, {$name}!", { name: string }),
  unread: message(
    `.input {$count :number}
.match $count
0   {{No unread messages.}}
one {{You have {$count} unread message.}}
*   {{You have {$count} unread messages.}}`,
    { count: number },
  ),
  cartEmpty: message("Your cart is empty.", {}),
};

const en = defineCatalogue("en-US", messages);

en.t("greeting", { name: "Ada" }); // "Hello, Ada!"
en.t("unread", { count: 0 });      // "No unread messages."
en.t("unread", { count: 1 });      // "You have 1 unread message."
en.t("unread", { count: 1200 });   // "You have 1,200 unread messages."
```

And a translation, negotiated and loaded lazily so a page ships one locale:

```js
const locales = defineLocales(en, {
  ja: () => import("./ja.js").then((module) => module.default),
});

const wanted = negotiate(
  parseAcceptLanguage(request.headers.get("accept-language") ?? ""),
  locales.available,
  "en-US",
);
const { t } = await locales.load(wanted);
```

## The five calls that do not compile

Run against a scratch file with `uf check`, verbatim (trimmed to the error lines):

```
tests/library/i18n-type-probe.js  6 errors
  error[incompatible-type]: Cannot call catalogue.t with object literal bound to args
    because property count is missing in object literal [1] but exists in ParamArgs [2].
      --> catalogue.t("unread", {});
  error[incompatible-type]: Cannot call catalogue.t with object literal bound to args
    because in property count: "12" [1] is incompatible with number [2].
      --> catalogue.t("unread", { count: "12" });
  error[prop-missing]: Cannot call catalogue.t because property unreadd
    (did you mean unread?) is missing in object literal [1].
      --> catalogue.t("unreadd", { count: 1 });
  error[prop-missing]: Cannot call catalogue.t because in type argument TKey: property
    unreadd (did you mean unread?) is missing in object literal [1].
  error[incompatible-type]: Cannot call catalogue.t with object literal bound to args
    because property count is extra in object literal [1] but missing in ParamArgs [2].
    Exact objects do not accept extra props.
      --> catalogue.t("greeting", { name: "Ada", count: 1 });
  error[incompatible-type]: Cannot call catalogue.t with object literal bound to args
    because property name is extra in object literal [1] but missing in ParamArgs [2].
    Exact objects do not accept extra props.
      --> catalogue.t("cartEmpty", { name: "Ada" });
```

The positive controls in the same file — `catalogue.t("greeting", { name: "Ada" })` and `catalogue.t("unread", { count: 3 })`, both bound to an annotated `string` — produce no errors. The probe file is not committed; `tests/library/i18n.test.js` asserts the positive half (every `t` call in it is annotated, so an argument type that degraded to `any` would show up), and the negative half has no home because `tests/type-tests/` is outside the directories this change may touch. Follow-up below.

## Why the parameters are values, not a type argument

Flow has no template-literal types, so `{ name: string }` cannot be derived from `"Hello, {$name}!"`. It has to be written beside it — and *how* is the whole design.

`message<{ name: string }>("Hello, {$nom}!")` would have been less to write and would have checked nothing: a type argument is erased before anything could look at it. Because the parameters are values, `message` gets the source and the parameters at run time and compares them:

```js
message("Hello, {$nom}!", { name: string });
// reads $nom, which is not declared in its parameters ($name)
message("Hello!", { name: string });
// declares $name, which the message never reads
message("Sorted by {$at :date}", { at: number });
// applies :date to $at, which is declared number; number takes :number or :integer
```

None of those is a type error in any checker. `translate` runs the same comparison for each locale against the source message's parameters, so a translator who drops a `{$count}` fails at start-up with the key named.

| | |
| --- | --- |
| a key that does not exist | Flow, at the call |
| a missing or misspelled argument | Flow, at the call |
| an argument of the wrong type | Flow, at the call |
| a message reading a parameter nobody declared | `message`, where it is written |
| a parameter the message never reads | `message`, where it is written |
| a parameter annotated as the wrong kind | `message`, where it is written |
| a translation that lost or invented a placeholder | `translate`, at start-up |
| a translation key the source locale does not have | `translate`, at start-up |
| a key no locale has translated | reported as `untranslated`, never a throw |

## The MF2 subset

The line is **the whole required default function registry, none of the draft one** — so a function uf accepts is one every conforming implementation must also accept.

**Implemented**

- Simple messages, quoted patterns (`{{…}}`), the four escapes `\\ \{ \} \|`.
- Variable placeholders `{$name}`, literal placeholders `{42}` and `{|two words|}`.
- `:string`, `:number`, `:integer`, `:date`, `:time`, `:datetime` with their options, including option values that are themselves variables (`{$n :number minimumFractionDigits=$places}`).
- `.input` and `.local` declarations, with an annotation put on a name by a declaration inherited by every later use of it.
- `.match` with any number of selectors; literal and `*` keys; exact numeric keys preferred over the plural category; MF2's variant sort, which a single scan for the "best" row gets wrong in exactly the two-selector messages nobody writes a test for.

**Not implemented, and refused where the message is declared rather than ignored**

| | Why |
| --- | --- |
| Markup, `{#bold}…{/bold}` | `t` returns a string. Dropping the tags loses a translator's emphasis silently; inlining HTML puts unescaped translator input into a page. A parts-returning API is where this belongs, and that is a different return type, not a bigger parser |
| Attributes, `{$x @unit}` | Conforming to accept and ignore, since they do not affect formatting. Refused because the only thing an attribute is for is a tool that reads it, and uf has none |
| `:currency`, `:unit`, `:math` | The draft registry. `:currency` is the one that will be missed; it needs a currency code in the message, which is a decision about where currency codes live |
| Reserved and private-use sigils, `{$x !foo}`, `{$x ^bar}` | Keeps a message that parses here from meaning something else under a conforming implementation later |
| Namespaced functions, `{$x :acme:money}` | Same reason |
| Bidi isolation **on** by default | The specification allows `bidiIsolation: none`; uf defaults to it because output goes into React children where the DOM already isolates, and two invisible code points per placeholder would make every string assertion in every application a puzzle. `defineCatalogue(…, { bidiIsolation: true })` turns it on, and both halves are asserted so the departure stays a decision |

MF2 decides simple against complex on the first character alone, and uf does not trim: a multi-line message begins with `.input` or `.match` hard against the backtick. A leading space makes it text, which is asserted rather than only documented.

## `Intl.MessageFormat`: uf formats MF2 itself

It is a TC39 proposal with no implementation in any shipping runtime, so building on it means building on nothing — and feature-detecting it with a fallback would be worse: two code paths deciding what a user reads, one of which has never run, and the same catalogue rendering differently in Safari and in Node.

So the algorithm is uf's and the data is `Intl`'s: `Intl.PluralRules` for categories, `Intl.NumberFormat` for numerals, `Intl.DateTimeFormat` for dates. The selection algorithm is a few hundred lines; CLDR's rules for the locales a browser already carries are megabytes. A `.match` on a number in a runtime without `Intl.PluralRules` raises rather than guessing English, because a message that silently pluralises Polish as if it were English is a bug that reaches production looking like a translation mistake.

The three constructors are declared narrowly in `format.js` rather than taken from the vendored `intl.js`, for the reason `@uniflowed/hooks`'s `timing.js` declares `Intl.RelativeTimeFormat`: the libdef predates `dateStyle` and `timeStyle` — exactly what MF2 defines `:date` and `:time` in terms of — and its option types have invariant optional properties, so an options object built at run time from a message cannot be passed to them at all. The local declarations are **stricter** than the libdef's, not looser: `:number`'s `style` is `decimal` or `percent` and nothing else, and each option is narrowed one at a time so a value MF2 does not define is dropped here rather than reaching `Intl`.

## Where it lives

A package of its own. `packages/std` is declaration-only native bindings whose functions throw `nativeRuntimeRequired`; this is ordinary Flow with no native binding and no dependencies, so it sits beside `@uniflowed/validator` and `@uniflowed/immer`. Five modules, each reachable through a subpath, nothing under `internal/`:

- `syntax.js` — MF2 source into a tree, and the refusals that define the subset
- `format.js` — the tree, a locale and arguments into a string
- `catalogue.js` — the types, the contract check, translations, lazy locales
- `negotiate.js` — `Accept-Language`, and RFC 4647 Lookup
- `index.js` — the barrel

`negotiate` has one documented departure from RFC 4647, and it runs last so it cannot outrank a real match: after every requested tag has failed, a pass matches on the primary language subtag alone. Lookup only shortens the *request*, so a reader asking for `en` does not match a catalogue that only has `en-US` — and the alternative for that reader is not British English, it is whatever the fallback happens to be.

## The `crates/` change, applied

`crates/uf_lib/src/registry.rs` mirrors every package's export list, and
`@uniflowed/i18n` is now in it: the sixteen values `packages/i18n/index.js`
exports, `NativeModuleKind::Data`, `Stability::Experimental`, sorted the way
`the_registry_names_exactly_what_each_package_exports` sorts them. Nothing
failed without it — that test walks the registry rather than `packages/` — but
`uf lint`'s `flow/native-module` rule and `uf dev`'s hover did not know the
package existed.

`tools/release/pending-packages.txt` names it too, which is what `Metadata` was
failing on. `tools/ci/publishable.sh` states the rule from the side that cannot
be forgotten: a package that never calls `nativeRuntimeRequired` is a real
implementation and must be named in one of the two release manifests, and this
one is ordinary Flow with no native binding. The pending list is the correct
half — `npm trust` cannot bind a name the registry has never seen, so putting
it in `published-packages.txt` would fail `publish.yml` after the names before
it had gone out. The closure check reads the published list only, so nothing
else moves.

The structural invariants in `crates/uf_lib/tests/package_surface.rs` are
checked by `cargo test --workspace`, which passes.

## Verification

Taken as a before/after against an untouched tree.

```
$ uf fmt --check

$ uf lint
  files checked  384
  errors         0
  warnings       14        # baseline: 14, none in the new files

$ uf test tests/library
✓ 2009 passed, 0 failed, 1 skipped in 7.64s     # baseline: 1951 passed, 0 failed, 1 skipped
                                                # 51 files; baseline 50. The 58 new tests are this change's

$ uf check
error: uf check failed with 1883 errors         # baseline: 1883 errors
```

`uf check`'s absolute count is not stable run to run — a per-file inference budget means a loaded machine gets a different number — so the comparison that matters is per file: **no file has errors after this change that did not have them before**, and `uf check packages/i18n tests/library/i18n.test.js` on its own reports `✓ no problems`.

## Worth a follow-up issue

1. **Build-time catalogue extraction.** `uf build` does not walk a project for `message(…)` calls, so there is no `messages.json` for a translation vendor and no build-time report of a key no locale translates. `messageUsage` is the reader; the walk is Rust's, under the same rule that puts the formatter and the checker there.
2. **A lint rule linking a message to its parameters.** The run-time check in `message` fires on the first import — a test, or the first render. A `uf lint` rule reading the string literal and the parameter object together would catch it in the editor, and uf already parses Flow in Rust, so it is a rule rather than a research problem.
3. **Markup, as parts.** `formatToParts` returning `$ReadOnlyArray<MessagePart>` would make `{#bold}…{/bold}` renderable as React elements without ever building an HTML string.
4. **`:currency`.** The one draft-registry function that will be missed.
5. **Publishing.** `@uniflowed/i18n` waits in `tools/release/pending-packages.txt` rather than `published-packages.txt`, because the other half of adding a name to the published list — `npm login`, `bootstrap-publish.sh`, `trust-npm.sh` — is a step a checkout cannot take. Moving it across is one line once that binding exists. Nothing published depends on it, so the closure check is unaffected.
6. **Negative type tests.** The five non-compiling calls above have no committed home — `tests/type-tests/` is where `@uniflowed/form` keeps its equivalents, and it is outside the directories this change may touch.

Closes #530.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/567"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791372458&installation_model_id=422833&pr_number=567&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F567&signature=a016c774c9cddb0419a953d50dbf5911e45f58d0c79397f2f41ad8939fea5344"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added the experimental `@uniflowed/i18n` package.
  - Supports typed MessageFormat 2 messages, parsing, formatting, translations, locale loading, and Accept-Language negotiation.
  - Provides locale-aware number, date, time, plural, and string formatting with validation and fallback handling.

- **Documentation**
  - Added an internationalisation reference guide and package reference entry.
  - Updated implementation status information.

- **Tests**
  - Added coverage for parsing, formatting, type validation, translations, locale loading, and negotiation.


ubugeeei:
Applied the `registry.rs` block — sixteen names, `cargo test -p uf_lib` 9 + 32 passed. Auto-merge armed.

The values-over-type-argument decision is the right one and the reason is the one that matters: `message()` refuses a mismatch **where it is written**, at declaration, rather than trusting a signature nothing checks at run time. A type argument would have been shorter and checked nothing.

Two notes:

- Leaving `@uniflowed/i18n` out of `published-packages.txt` was correct. An unbound name fails the publish job *after* earlier packages have gone out, which is a half-published release — the worst failure mode the pipeline has.
- `package-lock.json` was unavoidable and needed no apology: a workspace package that is not in the lock does not link under `npm ci`.

Your point 5 is a real gap and I am filing it — five negative type cases verified on a scratch probe and then thrown away is exactly the evidence that should have a committed home in `tests/type-tests/`.

PR #571, opened 2026-09-07
fix(react-testing, test): an accessible name, a summary, and two surfaces that were any

### LINKED ISSUE #375: react-testing: `accessibleName` puts `aria-label` before `aria-labelledby`, and gives a nameless table its cells
Two things `accessibleName` (`packages/react-testing/internal/queries.js`) gets wrong that #359 did not cover. Both were seen while adding the `<caption>` / `<legend>` / `<figcaption>` rule, and neither is what that issue asked for, so neither was changed.

## The order is inverted

```js
export function accessibleName(element: Element): string {
  const label = element.getAttribute("aria-label");
  if (label != null && label !== "") {
    return normalize(label);
  }

  const labelledBy = element.getAttribute("aria-labelledby");
```

The accessible name computation (`accname`) puts `aria-labelledby` **first**: an element carrying both is named by what `aria-labelledby` points at, and `aria-label` is the fallback. Here it is the other way round, so

```jsx
<button aria-label="Close" aria-labelledby="title">…</button>
```

is announced by every browser as whatever `#title` says and found by this query as "Close". Nothing in the repository writes both at once, which is why it has never bitten — a component that does would have a test that passes and a reader who hears something else.

## A table with no caption is named by its cells

`accessibleName` falls back to the element's own text for anything, and `table`, `figure` and `fieldset` are not "name from content" roles. So a captionless `<table>` has an accessible name of every cell in it — which is what made `getByRole("table", { name })` useless before #359, and is still true for the table that has nothing to be named by. HTML-AAM's next step after the caption is the `title` attribute and then *no name at all*; a query for `{ name: "" }` should find it and a query for its contents should not.

Fixing the second means deciding what the fallback is for every role rather than for elements, which is why it is its own issue rather than a line in #359.

### LINKED ISSUE #376: test: `toBeVisible` says a `<summary>` inside a closed `<details>` is invisible
`isVisible` in `packages/test/internal/expect.js`:

```js
    if (current.tagName === "DETAILS" && !current.hasAttribute("open") && current !== node) {
      return false;
    }
```

The `current !== node` exempts the `<details>` element itself from its own rule, and nothing exempts the `<summary>`. A closed disclosure renders exactly one thing — its summary — so

```jsx
<details>
  <summary>More</summary>
  <p>the rest</p>
</details>
```

`expect(screen.getByText("More")).toBeVisible()` fails, for the one element on the screen. The reader is looking at it.

The walk that `packages/react-testing/internal/queries.js` grew for ubugeeei-prod/uf#323 answers a different question — whether the accessibility tree exposes the element — and had the same decision to make, so there is a worked version of the rule beside this one:

```js
    if (
      child != null &&
      current.tagName.toLowerCase() === "details" &&
      !current.hasAttribute("open") &&
      child.tagName.toLowerCase() !== "summary"
    ) {
      return false;
    }
```

Carrying the child the walk came from is what makes "everything under a closed `<details>` except its summary" sayable. The two walks stay separate — they disagree about `opacity: 0` on purpose, and that is written down at both — but this half of the rule is the same half and only one of them has it.

### LINKED ISSUE #401: react-testing: `fireEvent` is published as `any`, and a proxy is why
`@uniflowed/react-testing` exports `fireEvent` with the type `any`:

```js
export const fireEvent: any = new Proxy(
  (target: EventTarget, name: string, init?: EventInit) => dispatch(target, name, init),
  { get(base, property) { /* one function per event name */ } },
);
```

A consumer who writes `fireEvent.clcik(button)` gets no error, and neither does one who passes an element where the library wants an `EventTarget`. That is a hole in the published type of the package whose purpose is testing *typed* components — the same argument `#381` made about the queries, which is why `screen`'s thirty-six names are written out today.

`packages/react-testing/internal/events.js` carries the two attempts and why each fails, and neither is a spelling problem:

- **An indexer.** `{ (target, name, init?): boolean, readonly [string]: (target, init?) => boolean }`. Flow declines it, and is right to: the value is a function, a function has `name`, `length`, `call`, `apply` and `bind`, and the `get` trap hands those back as themselves because `property in base` is true for them. None of the five is a DOM event, so the lie is unreachable from a real call — but a type is not something to be right about on average.
- **A written-out table.** `error[incompatible-type]: Functions without statics are not compatible with objects.` A `Proxy` over a function *is* a function, and Flow will not treat one with no statics as an object with properties, whatever those properties are.

So no type can be assigned to the value as it stands. Typing `fireEvent` means changing what it is: a function carrying real static properties, one per event name — which is a table of a hundred-odd entries that stops answering to the hundred-and-first, and that is a decision about a published API rather than a cast to remove in passing. The questions it turns on:

- Is `fireEvent.somethingNobodyListedYet(el)` a call this library wants to keep working? The proxy exists because the set of DOM events is long and grows.
- Testing Library's own answer is the table — `fireEvent` there is generated from a fixed event map. A table here would be the same shape and would type.
- `fireEvent(target, name, init)` already exists for a computed name, which is the escape hatch a table would need.

Until then the annotation is suppressed by name rather than renamed to `$FlowFixMe`, so it stays greppable and `uniflowed/unknown-lint-suppression` keeps the directive honest. It is one of seven `flow/unclear-type` suppressions in the packages; `#225` counts them.

### LINKED ISSUE #402: test: `expect` has no type, and the matcher interface is what would give it one
`@uniflowed/test`'s `expect` is `$FlowFixMe`, and so is everything it hands back:

```js
export const expect: $FlowFixMe = Object.assign(expectValue, { any: asymmetric.any, … });
function expectValue(received: mixed): $FlowFixMe { … }
function verdicts(received: mixed): { readonly [string]: (...args: $ReadOnlyArray<any>) => Verdict } { … }
```

So `expect(user).toBaa(1)` is not a misspelling anybody's checker will find, `expect(list).toHaveLength("3")` is not a type error, and `expect(p).resolves` on a value that is not a promise is fine until it runs. Every test in this repository is written against that, which is the same hole `#381` closed in `@uniflowed/react-testing`'s queries and `#242` closed in its render helpers.

The indexer in `verdicts` is where it starts. The entries do not agree about their arguments — `toBe` takes a `mixed`, `toHaveLength` a `number`, `toBeCloseTo` two — and `bind` applies whichever one was asked for to a `$ReadOnlyArray<mixed>` collected from the caller. Parameters are contravariant, so one indexer cannot describe both ends:

- `(...args: $ReadOnlyArray<mixed>)` rejects every entry that wants a `number`;
- `(...args: $ReadOnlyArray<empty>)` accepts every entry and rejects every call.

Narrowing it in place is therefore not available, and `mixed` with a cast at the call moves the same unsoundness one line without checking anything, because the caller's result is `$FlowFixMe` and its result's result is `expect`, also `$FlowFixMe`.

What makes any of it checked is the thing the code comment has been waiting for: **a written-out matcher interface** — one signature per matcher, plus `.not`, `.resolves` and `.rejects`, and the generic parameter that carries the received value's type through them so `expect("a").toHaveLength(1)` can be the error it is. That is a listing, the way `Queries` in `react-testing/internal/screen.js` is a listing: Flow has no template literal types, so the names have to be written for the type to exist at all, and the behaviour stays in one place.

Until then the `any` in the indexer is suppressed by name — one of seven `flow/unclear-type` suppressions in the packages, which `#225` counts.

ubugeeei:
Four correctness bugs in `packages/react-testing` and the assertion surface beside it. Two are answers a query gave that a reader would have disagreed with; two are published types that were `any`, and they are the same shape of problem — a value that answers to a name decided while the program runs.

## `accessibleName` had the accname order inverted (#375)

`aria-label` was checked before `aria-labelledby`. The spec has it the other way round: an element carrying both is named by what `aria-labelledby` points at, and `aria-label` is the fallback for when it points at nothing. So

```jsx
<button aria-label="Close" aria-labelledby="dialog-title">×</button>
```

was announced by every browser as whatever `#dialog-title` says and found by this query as "Close".

"Points at nothing" is two cases and both fall through now: an id that resolves to no element, and one that resolves to an element with no text. The second is what makes reordering *safe* rather than only correct — without it a label pointing at an empty span would name the button `""` and never reach the `aria-label` underneath.

## And it named a captionless table by its cells (#375, second half)

`textOf(element)` was the fallback for anything, and most elements are not named by their contents. A `<table>` with no `<caption>` was called `"Ada Lovelace 1815"`, so a query for its contents found it and a query for `{ name: "" }` — the question "is this the table with no name" — found nothing.

The fallback is decided per role now, which is what the issue said fixing it would take. `NAME_FROM_CONTENT` is ARIA's "name from: author, contents" list — eighteen roles, the things a person presses or chooses. For everything else HTML-AAM's next step is the `title` attribute and the one after that is no name at all. A role uf has no mapping for is not in the set either, which is the same answer for the same reason: `roleOf` returns `null` for `<div>`, `<fieldset>` and `<figure>`, and none of those is named by what it holds — the two HTML *does* name from inside are already answered by `namingChild`.

Nothing in the suite regressed: the roles queried by name across this repository are either in the list or named by `aria-label`, `aria-labelledby` or a caption.

## `toBeVisible` called a `<summary>` inside a closed `<details>` invisible (#376)

```js
if (current.tagName === "DETAILS" && !current.hasAttribute("open") && current !== node) {
```

The `current !== node` exempts the disclosure from its own rule, and nothing exempted the summary. A closed disclosure renders exactly one thing, so `expect(screen.getByText("More")).toBeVisible()` failed for the one element on the screen.

Saying "everything under a closed `<details>` except its summary" needs the child the walk arrived from, which is what `internal/queries.js`'s `exposed` already carries. The two walks stay separate — they disagree about `opacity: 0` on purpose — and that is now written down at both rather than at one.

## `fireEvent` was `any`, and a proxy was why (#401)

Both shapes the issue names really do fail, and neither is a spelling problem: an indexer beside the call signature is a lie about `name`, `length`, `call`, `apply` and `bind`, which the trap handed back as themselves; and a written-out table over a `Proxy` fails earlier still, because a proxy over a function *is* a function and Flow will not treat one with no statics as an object with properties.

So the value changed rather than the annotation. `fireEvent` is a real function carrying 105 real properties, attached inside a builder — the way `@uniflowed/test`'s `describe` attaches its modifiers, so importing the module still only declares — and `FireEvent` is the same list written down. The names are React's and Testing Library's, so a suite being ported already has these spellings.

## `expect` had no type, and the matcher interface is what gave it one (#402)

`Matchers<R>` writes out the forty-one names, one signature per matcher, plus `.not`; `Expectation` is that intersected with `.resolves` and `.rejects`; `Expect` is the callable with the asymmetric matchers on it. Generic in what a matcher *returns*, which is what lets `.resolves` reuse the one listing rather than be a second copy of it — and an intersection rather than an object spread, because a spread of an object type drops `readonly` off every property it carries over.

`Object.assign(expectValue, { … })` went with it. Flow models `Object.assign` as returning the *target*, so it produced a function with no matchers on it: eight `prop-missing` errors and a `flow/unsafe-object-assign` suppression sitting on top of them. Attaching the statics to a local before returning it is the same runtime value and types.

## What a typed `expect`/`fireEvent` cost

**The received value's type is not carried through the matchers.** This is the one thing #402 asked for that is not here, and it is written down in `internal/expect.js` and in the fixture rather than left to be rediscovered:

- `toBe` is `Object.is`, and identity is not assignability. `(expected: T)` demands the expected value be a subtype of the received one — a direction the runtime has no opinion about. `expect(document.activeElement).toBe(screen.getByRole("button"))` compares an `HTMLElement | null` with an `Element`, neither is the other's subtype, and that is the most common assertion in a DOM test. Jest and Vitest both type this argument as `unknown` for the same reason.
- Worse, the parameter has to be **inferred**, and `expect(x)` is where a lot of otherwise unconstrained expressions sit. `await expect(client.request("/users/1")).resolves.toEqual({ id: 1 })` stops checking and starts reporting that `request`'s own type parameter is underconstrained; `expect([])` becomes "cannot determine type of empty array literal". Measured rather than guessed: it was implemented first, and `uf check` reported 18 new errors across this repository's tests, of which about fifteen were correct assertions.

Four of those shapes are in `tests/type-tests/matchers.js` **unmarked**, so a future attempt fails at that file rather than in review.

**`expect(5).resolves` still types**, and still fails when it runs. It needs two call signatures on `expect` — one for a promise, one for everything else — and Flow then requires the single function behind them to satisfy both, which no single function does.

**`fireEvent` lost the spellings a proxy gave away for free.** `fireEvent.keydown` — the DOM's own lower-case name, which worked because the proxy lower-cased whatever it was handed — is gone, and so is any name off the list. `fireEvent(target, name, init)` is the escape hatch and always was; it is what the proxy called into, and `dispatch` is exported for the same reason. Both losses are marked in the fixture so they stay decisions rather than discoveries.

**Suppressions: two go, one arrives.** `flow/unclear-type` on `fireEvent`'s annotation and `flow/unsafe-object-assign` on `expect`'s are both gone. The new one is not an `any` at all: `flow/unclear-type` reads source text, and the shape `names_a_value` recognises as a property key rather than a type is a name at the start of a line or straight after `{`, `,` or `;`. `readonly any:` is neither, so the rule reports Jest's, Vitest's and Sinon's name for the matcher as an `any` type. The rule's own comment already lists `@uniflowed/test`'s `expect.any` among the false positives it exists to avoid; this is the one spelling it still cannot see past. Worth a follow-up on the rule, not a contortion in the type.

The one `flow/unclear-type` inside `verdicts` stays. Its indexer is now reached only between two functions in one file — a caller meets `Expectation` — and narrowing it means writing `bind`'s forty-one wrappers out by hand, which is a second copy of the listing to keep in step with the first.

## Verification

```
uf lint          0 errors, 14 warnings   (unchanged)
uf fmt --check   clean
uf test tests/library
                 before  1951 passed, 0 failed, 1 skipped
                 after   1961 passed, 0 failed, 1 skipped
```

`uf check` on an untouched tree already fails; before/after over the whole repository:

```
before  1883 errors, 14 warnings
after   1905 errors, 14 warnings
```

Every one of the 22 is accounted for:

| | |
| --- | --- |
| −8 | the `Object.assign` errors in `internal/expect.js` |
| +14 | the two new `tests/type-tests` fixtures, which exist to fail and are held to their exact diagnostics |
| +13 | `fireEvent.keyDown(document.body, …)` in `ui.test.js` and `hooks.test.js` — `document.body` is `HTMLElement \| null`, and the typed target says so |
| +3 | empty array literals inside `expect(…)` in `effect.test.js`, now visible because `expect` has a type at all |

`docs/roadmap.md` carried the 1,883 as a stated number, so it is updated.

## Tests

Eight behaviour tests in `tests/library/react-testing.test.js` — the `aria-labelledby` order and both of its fall-throughs, a captionless table, the `title` step, the roles that *are* named by content, and a summary visible inside a closed details with the rest of it not.

Two type tests, which is the interesting evidence for #401 and #402: a listing's claim is not provable by running anything, because no assertion about behaviour can say a *different* program would have been rejected. `tests/type-tests/matchers.js` and `tests/type-tests/event-names.js` are the misuses written down; `tests/library/assertion-types.test.js` runs `uf check tests/type-tests packages/test packages/react-testing --json` and holds each file to its own `// expect:` markers — a marked line that stops being an error fails, and so does an unmarked line that starts being one.

## `crates/` diff needed: none

`crates/uf_lib/src/registry.rs` mirrors each package's **value** exports, and `the_registry_names_exactly_what_each_package_exports` compares them. Everything added here is a type — `Expect`, `Expectation`, `Matchers`, `FireEvent`, `Firer`, `EventInit` — and that test's own comment says why types are not listed: a type is imported with `import type`, which the checker resolves through the file itself. No registry change is needed, and none was made.

## Worth a follow-up

- **`flow/unclear-type` cannot see past `readonly` in front of a property key.** `names_a_value` in `crates/uf_lint/src/runner/flow_type.rs` accepts `{`, `,`, `;` or the start of a line before a key; the modifier defeats it. One suppression here exists only for that.
- **13 call sites pass `document.body` to `fireEvent`**, in `ui.test.js` and `hooks.test.js`. It is `HTMLElement | null`; the typed `fireEvent` now says so, and the tests should narrow rather than the type widening.
- **`packages/testing/index.js` re-exports a type `Screen` that `@uniflowed/react-testing` does not export.** Pre-existing — it is on the baseline too — and the export is `Queries`.

Closes #375.
Closes #376.
Closes #401.
Closes #402.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/571"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791372679&installation_model_id=422833&pr_number=571&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F571&signature=736af1a455c8f1102e1fd5f04bb83423fa7a971b409b6e23e79241c527f03e2c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added typed public APIs for test matchers and React event helpers, improving editor guidance and catching misspelled names or invalid arguments.
  * `fireEvent` now supports explicit, camel-cased event helpers while retaining dynamic event-name calls.
  * Improved accessible-name calculation, including `aria-labelledby`, `aria-label`, content-based names, and `title` fallback behavior.
  * Corrected visibility handling for content inside open and closed `<details>` elements.

* **Documentation**
  * Updated documented type-check results and clarified intentional diagnostics.


ubugeeei:
The suppression you had to write is a linter bug, and it is fixed in [#572](https://github.com/ubugeeei-prod/uf/pull/572): `flow/unclear-type` recognised `{ any: … }` as a property key and `{ readonly any: … }` as a type annotation. Six spellings of the same key now read as keys — bare, `readonly`, `+`, `-`, `?`, and `readonly …?` — with a test each in the other direction so a real `any` beside one is still reported.

Once #572 lands, please drop the `flow/unclear-type` suppression from this branch. A suppression that exists because the rule is wrong is exactly the escape hatch this repository does not want.

The rest of the report is excellent — particularly measuring the typed-received-value variant and *not* shipping it. Eighteen new errors, fifteen of them on correct assertions, with `expect([])` becoming "cannot determine type of empty array literal", is a decisive result; Jest and Vitest typing it `unknown` for the same reason is the confirmation. Four unmarked shapes in the fixture so a retry fails there is the right way to leave it.

Filing your points 2 and 3.

Issue #575, opened 2026-09-07
config: `lint.ignore` is read by fmt, check, test and doc, and its name says otherwise


ubugeeei:
`is_ignored` in `crates/uf_project` reads `config.lint.ignore`, and every command that walks the project goes through it: `uf fmt`, `uf lint`, `uf check`, `uf test`, `uf doc`.

So the one list is named after one of the five commands that read it. A person excluding a directory from formatting has to put it under `lint`, and a person reading the configuration reference has no reason to think `lint.ignore` does anything to `uf fmt`.

Honouring `.gitignore` ([#483](https://github.com/ubugeeei-prod/uf/issues/483)) takes most of the weight off it — the common case is a generated file, and that is now covered without any configuration at all — but the key is still there and still misnamed.

**Done**: the project-wide list has a project-wide name, `lint.ignore` keeps working as an alias for as long as alpha lasts with a note saying which it is, and the configuration reference stops needing a paragraph to explain that a key does not mean what it says.

Worth deciding at the same time whether a per-command ignore is wanted at all. It may not be: after `.gitignore`, the remaining cases are "this directory is not source", which is not a per-command fact.

Issue #564, opened 2026-09-07
check: the same tree reports different error counts depending on cache state


ubugeeei:
`uf check` over an unchanged tree reported **3439** and **1883** errors an hour apart, and a warm cache disagreed with a cold one by roughly 1600 on the same tree. Only a back-to-back cold pair is comparable, which means the number is not usable as a number — you cannot say whether a change made things better or worse without deleting the cache first and knowing to.

`docs/security.md` claims check-cache entries are keyed so that they invalidate correctly. Either that is not true, or the entries are correct and something else in the run is order-dependent; either way the tool is reporting a fact about its own state rather than about the code.

**Done**: two cold runs over the same tree report the same number, a warm run reports the same number as a cold one, and there is a test that says so — the same tree, checked twice, with a cache in between. If the cause turns out to be a key that omits something the answer depends on, that key is the fix; if it is inference that depends on which files were already resolved, that is a different and larger finding and worth saying so out loud.

Until then, a number from `uf check` should not be quoted without saying whether the cache was cold — which is what every report in this repository has had to do, and is itself the evidence.

Found while verifying [#563](https://github.com/ubugeeei-prod/uf/pull/563).

ubugeeei:
**This is worse than a wrong number: it fails a test.**

While fixing #417, `module-mock.test.js > reports the wrong type for an export as an error…` went red on an unchanged tree and green again later — on `main` as well as on a branch — with `.uf/cache/check` as the only variable. Running the underlying `uf check` by hand, **cold**, gave the fixture zero diagnostics in one run and the expected `incompatible-type` in another.

So it is not only that the count is unquotable. A library test's result depends on cache state, which means a green suite is not evidence and a red one is not either. That is a ready-made reproduction and the place to start triage.

Raising this from "the number is not usable" to "the checker is not deterministic", which is a different severity.

PR #576, opened 2026-09-07
fix(project): a file git is told to ignore is not the project's to format

### LINKED ISSUE #483: fmt: `uf fmt` has no ignore list, so it formats generated files a `.gitignore` already excludes
`uf fmt` has no way to be told a file is not the project's to print, and `fmt` in the configuration reference has no `ignore`. It formats generated output as readily as source, including a directory whose own `.gitignore` is `*`.

Here the wasm-bindgen glue is generated by the Rust build into `src/wasm/`, which holds a `.gitignore` containing `*`:

```
$ uf fmt --check
  - src/wasm/glue.js
! 1 file of 258 needs formatting
```

So `uf fmt --check` fails on a clean checkout as soon as the generated file is built, `uf fmt` rewrites a file that the next build overwrites, and the only way out is to keep uf away from the whole tree.

`uf lint` does not report this file, so the two commands already disagree about what the project is; `lint.ignore` exists and `fmt` has no counterpart.

Either honouring `.gitignore` or accepting `fmt.ignore` would settle it — `lint.ignore` and `lint.files` reading for both would be one list rather than two.

uf 0.0.0-alpha.10, aarch64-macos.

ubugeeei:
Closes #483. Files #575.

`src/wasm/` holds wasm-bindgen glue and a `.gitignore` of `*`. `uf fmt --check` failed on a clean checkout as soon as it was built, `uf fmt` rewrote a file the next build overwrote, and the only way out was keeping uf away from the whole tree.

## The answer the issue preferred

> Either honouring `.gitignore` or accepting `fmt.ignore` would settle it — `lint.ignore` and `lint.files` reading for both would be one list rather than two.

`.gitignore` is the list the project already keeps of what is not its source, and a generated file is generated whichever command is asking. So the project walk reads it, and `uf fmt`, `uf lint`, `uf check`, `uf test` and `uf doc` agree by construction rather than by four configuration keys somebody has to hold in step.

`ignore = "0.4.25"` was **already a workspace dependency that nothing used**, so this costs no new crate.

## Three defaults turned off, each on purpose

- **The global gitignore.** A developer's own `~/.config/git/ignore` must not change what `uf fmt --check` says — two people on one repository have to get the same answer.
- **Parent directories.** A `.gitignore` above the project root belongs to whatever the project is sitting inside; a checkout in somebody's `~/ignored-scratch/` is still a project.
- **Hidden files.** uf walked them before this, and stopping would be a second change hiding inside the first.

Negations work (`!src/kept.js`), with a test, because a `.gitignore` is a program rather than a list and half-reading one is worse than not reading it.

## What I did *not* add

`fmt.ignore`. `lint.ignore` is already read by all five commands — it is the project-wide list wearing one command's name — so a second key would make it two misnamed lists instead of one. That is #575, and the configuration reference now says plainly what the key does rather than what it is called.

## Verified

Reproduced the issue's exact shape, and confirmed both directions:

```
$ uf fmt --check            # src/wasm/.gitignore is `*`

$ uf fmt --check            # after adding an unignored, badly formatted file
  - src/messy.js
! 1 file of 3 needs formatting
```

`cargo test -p uf_project` 37 passed (was 35), uf_cli lib 298, cli 66, every_command 13. `cargo fmt --check`, clippy and `uf fmt --check` clean. Against this workspace: fmt clean and lint at the same 14 warnings, so nothing that was being checked stopped being checked.


ubugeeei:
Pushed a second commit, and the failing `Library` job was right: this was my bug, not the #564 flake. I checked rather than assumed — `main` passed the same file, my branch failed it 3 out of 3.

`module-mock.test.js` writes a fixture into a **gitignored** directory (deliberately, so a killed run leaves nothing behind) and then asks `uf check` about it *by name*. Honouring the ignore answered "no diagnostics" for a file that has one — silence that reads as success, which is the exact failure shape this PR exists to prevent.

So the rule has a second half now, and it is the one every tool in this class already has: **naming a path suspends `.gitignore` for anything under it.** The ignore file says what your source *is*, not what you may ask about. `.uf` and `.git` are not suspended, and `lint.ignore` still applies.

Verified both directions on the reproduction: whole-project skips `src/wasm/glue.js`, `uf fmt --check src/wasm` reads it. Library suite back to 1,951 passed / 0 failed.

ubugeeei:
Second failure found and fixed, and it is the same lesson in a new place.

`uf prepare` writes `router.js` and the server-action types and then checks them — a generated file that does not compile is a broken build whether or not anybody staged it. They are also exactly the files a project tells git to ignore; `router.js` is in this repository's own `.gitignore`. So the new walk found neither, and `uf prepare` generated two files and silently checked nothing.

`an_empty_index_skips_the_checks_rather_than_widening_to_everything` and `the_generated_files_are_checked_even_though_they_are_never_staged` both went red, and they were right to.

Fixed by passing the generated paths as the named set — the rule from the previous commit, applied by the one caller that already knew which files it meant.

Full `cargo test -p uf_cli`: 24 suites green, 0 failures. `uf test tests/library` 1,951 passed. `cargo fmt --check` clean.

PR #582, opened 2026-09-07
feat(hooks, router): the address bar, and a hydration error that names the node

### LINKED ISSUE #508: dev: hydration errors that show the difference
React says "Hydration failed because the server rendered HTML did not match the client". uf can say more, because it has both trees and it owns the dev overlay.

**Done**: the overlay shows the server's markup and the client's at the node that differed, the component that rendered it, and — when it is one of the three usual causes — which one. Related: the three causes are #NEXT.

Part of #488.

### LINKED ISSUE #515: web: the browser APIs an application reaches for, typed and SSR-safe
`@uniflowed/hooks` has some of this. What an application still writes by hand is the pattern rather than the API: a browser-only value that must not run during a prerender, must not differ at hydration, and must be typed.

Clipboard, geolocation, media queries, intersection, resize, visibility, online status, storage, permissions, the URL bar.

**Done**: each is a hook with a server-safe default and a documented first-render value; each has a test that renders it in a prerender and asserts what the markup contains; and the ones that cannot have a server value say so in their type rather than returning `null` and hoping.

Part of #488.

ubugeeei:
Two issues that meet at one seam: what a component may read from the browser during a render, and what happens when one of them gets it wrong.

## #515 — the browser APIs an application reaches for

`@uniflowed/hooks` already answered nine of the ten APIs the issue names — clipboard, geolocation, media queries, intersection, resize, visibility, online status, storage, permissions. The tenth was the URL bar, and two of the nine answered with the shape the issue calls out by name: `null` and hoping.

### `useHash`

In `browser.js` rather than a module of its own, because the address bar is a reading of the one browser the page is in and that is what the file is for. It is also the only URL hook there will be: the path and the query are on the request, so `@uniflowed/router` has already resolved them and `useRoute` is where a page reads them. **The fragment is the one part of the URL nothing on the server has ever seen** — the browser strips it before the request goes out.

That fact shapes the signature. `useHash` takes no `serverValue` where `useMediaQuery` and `useOnline` do: `""` is not a default standing in for something better, it is what the server knows, and an override would only have invited somebody to state a value that cannot be true.

Three things it does that the hand-written version gets wrong:

- **It hears its own writes.** Neither `history.pushState` nor `replaceState` fires anything at all — not `hashchange`, not `popstate` — so a hook built only on the platform's events leaves every component except the one that wrote showing the fragment from before the write. A module-level registry, the same shape as `useStorage`'s and balanced under Strict Mode the same way, closes it. A `pushState` made anywhere else stays invisible, and the doc names the gap rather than pretending there is none.
- **It does not scroll.** Moving to a section and recording which tab is open are separate intentions; the platform couples them only because assigning `location.hash` is the old way of doing both at once.
- **It writes through `URL`.** See the security section below.

### The two that could not say what they did not know

`GeolocationReading` was `{| position: Geoposition | null, error: Error | null, supported: boolean |}` — eight representable states, four of them impossible, and `position == null` meaning "no browser", "not asked", "refused" and "still waiting" at once. It is now five disjoint states (`idle`, `unsupported`, `pending`, `located`, `failed`), and `enabled: false` reports `"idle"` *before* `"unsupported"` is considered, so the answer is the caller's own and the same on both sides of a hydration.

`Network` puts the three fields that come from `navigator.connection` behind one `measured: NetworkMeasurement | null`, and keeps `online` flat — it is the field almost every caller wants, and burying it behind a narrowing would make the common case pay for the rare one. `measured: null` now says exactly one thing.

**Breaking**, for two hooks `index.js` already lists as Experimental.

### Documented first-render value

`browser.js`'s header carries a table of what every hook there renders during a prerender, in the client's hydrating render, and afterwards. `tests/library/hooks-ssr.test.js` asserts every row, so one that stopped being true fails there rather than in somebody's browser — and that file now also covers the last seven names in the package, which were the only ones it had never rendered while `index.js` claimed the whole surface. `docs/app/reference/hooks` is the same table for a reader without the source open.

## #508 — hydration errors that show the difference

React reports a mismatch with one sentence and a list of six possible causes. The list is a good list and it is not an answer: the node that differed is still somewhere in a page of two thousand of them.

uf has both trees. `@uniflowed/router`'s client entry is the only code that runs between the parser finishing and `hydrateRoot` starting — which is the only moment the server's markup exists, because React repairs a mismatched subtree by rendering over it — so it takes a copy there, and `packages/router/internal/hydration.js` walks that copy against the live DOM when React reports:

```
Hydration mismatch in <Posted>

at       main:nth-child(1) > p:nth-child(2) > text()[0]
server   3 minutes ago
client   5 minutes ago

The two renders computed the same text from a value that is different every
time it is read — a clock, a random number, or a date formatted in a locale.
Decide the value once, above the tree, and pass it down, so both renders read
the same number instead of each working one out.

React said: Hydration failed because the server rendered text didn't match…
```

The node, the two values, the path and the component — the four things React does not say. The component is the innermost frame of React's stack that is not a host element, because the `<p>` React names is already in the path a line above.

Three causes are named where the difference gives them away — a value that moves, a reading only a browser can make, and markup that cannot nest — and anything else is reported as `unknown` rather than guessed at. A wrong diagnosis sends the reader to the wrong line and makes the whole panel less believable.

The same text goes to the console, so a headless run sees it. It is gated on `import.meta.hot`, the one signal that is right in all three places `client.js` is evaluated: Vite defines it while serving, replaces it with `undefined` in a build so the branch is statically dead, and Node leaves it undefined so `rsc-split.test.js` gets the production path. The import is dynamic as well, so a production bundle has no path to the module rather than merely no call into it.

**This is deliberately not the whole story.** #554's render envelope is what makes most of these not happen at all; this is for the ones that still do. And it does not reach the terminal: `uf dev`'s diagnostics come up the driver's event channel from the Node process, and a mismatch happens in the browser, which has no channel to send one back on. That is named in the module header rather than half-built — see the follow-ups below.

## Security

Checked against `docs/security.md`, row by row, for the four it touches:

- **"Decide on the canonical form"** (rule 2), against the fragment writer. The obvious spelling — `history.pushState(null, "", "#" + next)` — hands a caller's string to a URL parser, and `../admin?token=1` resolves to a different path with a query on it: a same-origin rewrite `pushState` permits, out of a call that promised a fragment. `writeFragment` builds `new URL(location.href)` and assigns `url.hash`, so the value can only ever land after the `#`, escaped. `tests/library/hooks.test.js` asserts the origin, the pathname and the search are unchanged rather than asserting the spelling.
- **"XSS via CSP nonce handling and script injection points are typed, not string-concatenated"** — the framework/RSC table's last row — against the overlay. It shows the page's own markup, which on a page whose mismatch is in somebody's comment is attacker-authored. Every value goes in through `textContent`; nothing is interpolated into HTML and there is no `dangerouslySetInnerHTML` anywhere in the change. The test asserts the consequence: `` appears as characters, no `` element exists in the panel, and the handler never ran.
- **"No unbounded anything"** (rule 4), against the snapshot and the walk. The server markup has a 512 KB ceiling above which nothing is kept and the report says so; the comparison has a node ceiling and a depth ceiling on an explicit stack rather than the call stack; each shown value is truncated. Every hydration on every page pays the snapshot, so the ceiling is a cost decision as well as a safety one.
- **"No regex on untrusted input"** (rule 5), against the classifier. Both sides of a difference and React's message embed the application's own rendered content. The variable-input test is a hand-written left-to-right digit scan and the nesting test is `String.includes`; there is no regular expression on any of them.

Also checked and not applicable: CVE-2025-29927's class (nothing here reads a request header — `useHash` reads a value the request never carried), the `server.fs.deny` bypasses and the dev-server rows (no new file serving, no new endpoint), and the image/font decoder rows.

## Verification

| | Before | After |
| --- | --- | --- |
| `uf test tests/library` | 1951 passed, 0 failed, 1 skipped, 50 files | 1985 passed, 0 failed, 1 skipped, 51 files |
| `uf lint` | 0 errors, 14 warnings | 0 errors, 14 warnings |
| `uf fmt --check` | clean | clean |

`uf build` in `docs/` also succeeds and prerenders the new reference page.

No `uf check` totals, because the count depends on cache state (#564) and I have no cold-cache pair to report. Per file, which does not: `packages/hooks/browser.js`, `packages/hooks/index.js`, `packages/router/client.js`, `packages/router/internal/hydration.js` and `tests/library/hydration.test.js` each report **0 errors**, and `tests/library/hooks-ssr.test.js` went from 13 to 11 (the two `useNetwork`/`useGeolocation` reads that no longer exist). `tests/library/hooks.test.js` is unchanged at 70, all of them pre-existing.

## One change to `crates/` I could not make

`crates/uf_lib/src/registry.rs` mirrors each package's root exports and a Rust test compares the two, so this needs the one new name added in both of its lists. Alphabetical position in each, `useHash` between `useGeolocation` and `useHover`:

```diff
@@ registry.rs, builtin_modules(), the "@uniflowed/hooks" entry @@
                 "useFocusWithin",
                 "useGeolocation",
+                "useHash",
                 "useHover",
                 "useIdle",

@@ registry.rs, hook_descriptors() @@
         HookDescriptor::new("useFocusWithin", true, false),
         HookDescriptor::new("useGeolocation", true, false),
+        HookDescriptor::new("useHash", true, false),
         HookDescriptor::new("useHover", true, false),
         HookDescriptor::new("useIdle", true, false),
```

`idempotent_render: true` — it is a `useSyncExternalStore` and writes nothing during a render. `server_component_safe: false`, for the reason the function's own doc gives about every hook but `useIsomorphicLayoutEffect`: it reports a stated server value that the client will disagree with on hydration, and a Server Component has no client.

No other name changed. The three new types are types, and that table is values only.

## Worth a follow-up

- **The overlay cannot reach the terminal.** `packages/hmr` has an `EventSource` client and a documented `/__uf/hmr` endpoint, and nothing serves it; `ConnectOptions` has no `onError` either. A browser-side diagnostic that appeared in the same terminal as every other uf diagnostic is the obvious next step and is a whole piece of work.
- **A fourth cause: a browser extension.** React names it and this cannot detect it, because an extension that mutated the DOM before `hydrateRoot` did so before the snapshot too. Detecting it would mean comparing against the bytes on the wire rather than against the parsed tree.
- **`useHash` and a `pushState` made elsewhere.** The Navigation API's `currententrychange` would cover it where it exists, including uf's own router's navigations. It is Chromium and recent Safari only, so it is an addition rather than the base, and it wants its own issue.
- **The router does not track the fragment.** It writes one during navigation and scrolls to it, and `RouteInfo` has no field for it — which is why `useHash` exists in the hooks package at all. Whether the fragment belongs in `useRoute` is a router question worth asking separately.

Closes #515.
Closes #508.


ubugeeei:
Applied both `registry.rs` edits — the export list and the hook descriptor, `idempotent_render: true` / `server_component_safe: false` as you specified. `cargo test -p uf_lib` 9 + 32 passed. Auto-merge armed.

Two things in this report are exactly right and worth saying so:

**Adding only `useHash`.** Nine of the issue's ten APIs already existed, and a second spelling of any of them would have been the wrong kind of thorough. Naming the gap and filling only that is the harder call.

**No `serverValue` on `useHash`'s snapshot.** `""` is not a default standing in for something better — the browser strips the fragment before the request goes out, so it is *what the server knows*. An override parameter would have invited a caller to state a value that cannot be true. That distinction is the whole difference between a correct SSR story and one that produces a mismatch on the first paint.

And declining to convert `useGeolocation` and `usePermission` to stores, because neither has a snapshot to read, is the right reading of the rule rather than a mechanical application of it.

Filing your follow-ups 1 and 3.

docs/app/reference/hooks/_uf.page.mdx
```diff
@@ -0,0 +1,170 @@
...
+```js
+// @flow
+const where = useGeolocation();
+switch (where.status) {
+  case "idle":        // enabled: false — nobody asked for a watch
+  case "unsupported": // a server render, or a browser without geolocation
+  case "pending":     // the reader has been asked and has not answered
+  case "located":     // where.position is a Geoposition, guaranteed
+  case "failed":      // where.error, and no earlier fix to fall back on
+}
```
ubugeeei: you must use `match`

ubugeeei: Done — the block is a `match` now, and so is the one other place in this diff that had the same shape spelled out in `if`s.

**`docs/app/reference/hooks/_uf.page.mdx`.** The `switch` was five empty `case` labels used as a list of the union's members: control flow that does nothing, and a shape nothing else in this repository is written in. It is a `match` over `where` now, with an arm per state — and it earns its place, because it demonstrates the sentence underneath it rather than restating it. `position` exists in the `"located"` arm's binding and nowhere else, and dropping an arm stops the file compiling instead of falling through to the next label.

```js
const label = match (where) {
  {status: "idle"} => "Not watching",
  {status: "unsupported"} => "This device cannot say",
  {status: "pending"} => "Locating…",
  {status: "located", position: const at} => `${at.latitude}, ${at.longitude}`,
  {status: "failed", error: const error} => error.message,
};
```

**`packages/router/internal/hydration.js`.** `classify` read `difference.kind` through five sequential `if`s, and two of the six kinds — `"node-type"` and `"tag"` — reached their answer by falling off the end of the function. All six are named in a `match` now, so a kind added to `DifferenceKind` stops this file compiling rather than arriving in somebody's overlay as `"unknown"`. Behaviour is unchanged; the two guards are `if (…)` arms so the ordering that was in the `if`s is still in the arms.

The arms that used to be implicit now have tests in `tests/library/hydration.test.js`: an attribute on one side only, an attribute whose value is not a moved digit, and two trees holding different nodes at the same position.

**Swept the rest of the diff, and there was nothing else.** `useGeolocation`'s own `useMemo` builds the union rather than reading it, and its `if`s are an ordered precedence over `enabled`, `supported` and `position != null` whose order carries the hydration argument written above it — a `match` over a tuple of three booleans would hide that, not express it. `firstDifference`'s `frame.kind === "unmatched"` is a single early return, not a dispatch. Everything else that looked like a candidate is a two-way ternary.

**Checked**, from a worktree with the pinned nightly, after merging `main` in: `cargo build --release --bin uf` clean, `uf fmt --check` clean, `uf lint` 0 errors / 14 pre-existing warnings over 401 files, `uf test#library` on node 24 — the two new cases pass, and the run is green apart from one flake that is not mine: `serve.test.js` and `standalone.test.js` both assert `state.pulls` on a socket that has just drained, and on a loaded machine the drain fires twice (`expected 2 to be 1`). Both live on `main` untouched by this branch, both pass on a re-run, and 0fd19a2 changes three files none of which is in `packages/server`. Nothing in Rust changed, so no `cargo test --workspace`.

0fd19a2.

Issue #405, opened 2026-09-06
server: a request Node's parser refuses is answered with 400 and logged nowhere


ubugeeei:
`packages/server/node.js` registers no `clientError` handler on its
`createServer`, so when Node's own parser refuses a request the server says
nothing at all — no stderr line, nothing in the request log. Node answers
`HTTP/1.1 400 Bad Request\r\nConnection: close\r\n\r\n` and `nodeListener`
never runs.

## How it was found

A test spent a day looking like a streaming bug. `assert_served` builds a URL
from its `command` parameter, and the node-adapter caller passes a whole
command *line*, so assertion 5 sent:

```
GET /slow/build --adapter node HTTP/1.1
```

A request target may not contain a space. llhttp rejects it with
`HPE_INVALID_CONSTANT`, `http.Server`'s default handler answers the bare 400,
and nothing uf wrote was involved. The test's message said

```
`uf build --adapter node` did not render the suspending route
HTTP/1.1 400 Bad Request
Connection: close
```

and the server's stderr said nothing, so the obvious readings were all wrong
ones: the streaming renderer, the adapter's handler, a desynchronised
connection.

The test now prints the request as it went on the wire and names the response
as Node's own (ubugeeei-prod/uf#353), which is what makes the *test* debuggable.
The server is still silent.

## Why this is worth its own change

An operator whose client sends a malformed request gets exactly the nothing
this investigation got. `uf start`, `uf preview` and every adapter share that
server, so it is the production answer for all of them.

It is a deliberate-looking gap with no comment recording a decision, which is
the shape this repository treats as a bug rather than a choice.

## What "done" means

- `createServer(...).on("clientError", …)` in `packages/server/node.js`, saying
  what was refused and why, in uf's own diagnostic voice rather than a stack.
- **The log-volume question answered in the comment**, because it is the reason
  someone might not want this: malformed requests are cheap to send, so a line
  per rejection is a denial-of-service amplifier on a public address. Rate
  limiting, sampling, or a count are all defensible; picking one silently is
  not.
- Node's own 400 must still be what goes on the wire — this is about saying so,
  not about answering differently.
- A test that does not need a socket. `crates/uf_cli/tests/vite.rs`'s
  `a_bare_400_says_it_is_node_refusing_the_request_line` shows the shape, and a
  `stream.Duplex` handed to `http.Server` as a connection drives the real
  parser without `listen`, which is how the diagnosis above was made on a
  machine where `bind` is refused.

PR #585, opened 2026-09-07
feat(server): OAuth as a contract, and a logger with a request id a render can read

### LINKED ISSUE #505: server: OAuth, as a contract rather than a provider
An application that signs a user in needs a redirect handler, a state parameter it can verify, a token exchange, a session, and a way to read the user in a loader and in a server component.

The red lines are what make this an issue rather than a package: "Keep provider SDKs and proprietary services out of portable application and framework contracts". So the deliverable is the contract — authorize, callback, refresh, session — with one implementation, not an integration with a company.

**Done**: the four route handlers uf generates, the session store behind an interface, PKCE by default, and a test that a callback with the wrong `state` is refused. See `docs/security.md`.

Part of #488.

### LINKED ISSUE #506: server: structured logging, with a request id that reaches a render
#405 is the symptom: a request Node's parser refuses is answered with 400 and logged nowhere. There is no logger, so there is nowhere for it to have gone.

**Done**: a structured logger with levels, a request id created at the edge of the server and readable from a loader, a route handler and a server component; JSON in production and a readable line in `uf dev`; and the one thing that makes it worth having — the log line for a request carries the route that matched, not only the path.

Part of #488, and the missing half of #405.

ubugeeei:
Two issues in `packages/server`, landed together because the second one uses the
first: the OAuth flow's refusals are the thing most worth having a log line for.

## #505 — the seam

uf ships no provider, and the contract is one type with four strings and one
function:

```js
export const example: OAuthProvider = {
  authorizationEndpoint: "https://example.com/oauth/authorize",
  tokenEndpoint: "https://example.com/oauth/token",
  clientId: process.env.EXAMPLE_CLIENT_ID ?? "",
  clientSecret: process.env.EXAMPLE_CLIENT_SECRET,
  scope: "openid email",
  async identify(tokens) {
    const answer = await fetch("https://example.com/userinfo", {
      headers: { authorization: `Bearer ${tokens.accessToken}` },
    });
    if (!answer.ok) throw new Error(`userinfo answered ${answer.status}`);
    const user = await answer.json();
    return { subject: String(user.id), claims: { email: user.email } };
  },
};
```

The provider in `tests/library/oauth.test.js` is exactly that, written from
nothing but the type — the evidence rather than the claim. `createAuth` turns it
into `authorize`, `callback`, `refresh` and `session` (`GET` who, `DELETE` sign
out), each a plain `Request` → `Response`, plus `currentSession()` and a
deliberately separate `tokens()`.

The store is four methods and one value type. Three are obvious; `take` is the
design — it reads *and* removes in one step, and that operation is what makes a
`state` parameter single-use. A read followed by a delete has a window between
them, and a window is all a replay needs.

**uf owns the flow, the state, PKCE, the cookies, the redirect, the session and
where the tokens are not. A provider owns two URLs, a client id, a scope and
`identify`.** There is no `providers/` directory and there will not be one.

## #506 — the request id

`requestId()` and `logger()` are readable from a loader, a route handler and a
server component's render, from an `AsyncLocalStorage` that already holds
`headers()` and `cookies()` — so two requests in flight see their own. Reading
the id counts as a read of request state, so a render that puts it in the
document is a render the route cache refuses to store; `logger()` deliberately
does not count, because an id that reached a log line has not made the document
personal.

Every request leaves one line, and it carries **the route pattern** rather than
only the path — `noteRoute` on the context, written by the router's dispatcher
and by its renderer through a new early `onMatch` on `resolveMatch`, because a
loader runs inside that call and a loader has things to log. `serve` now
attaches the `clientError` handler #405 is about.

## Security cases checked

Against `docs/security.md`, which gains a **Signing in** section and a
**Logging** section for these. The rows that applied:

| Rule or past failure | What it forced here |
| --- | --- |
| `Origin` against `Host`, never `X-Forwarded-Host` | `sameOrigin` compares the header against `new URL(request.url).host` and reads no forwarded header anywhere; a missing `Origin` is refused, not allowed |
| No request header steers control flow — CVE-2025-29927's class | Nothing in the flow branches on a header. The `redirect_uri` sent at exchange time comes from the stored pending record, so two requests with two `Host` headers cannot be made to disagree; `X-Request-Id` is not read at all |
| Nothing personalised becomes cacheable — the RSC cache-poisoning row | `Cache-Control: no-store, private` and `Vary: Cookie` on every auth response; `currentSession()` reads `cookies()`, so a page that asks who is signed in is never route-cached |
| Rule 2 — decide on the canonical form, then use the value you checked | The return path is validated when stored and again when used; the `redirect_uri` is stored rather than re-derived |
| Rule 3 — deny by default, allow by table | The return path is a closed shape, not a host list; the redaction table is names, not value heuristics; the token endpoint must be `https` or loopback |
| Rule 4 — no unbounded anything | 64 KiB ceiling on the token response with the stream cancelled above it; an entry ceiling on the in-memory store; depth, key, entry and string bounds on every log record |
| Rule 5 — no regex on untrusted input | None added. `pathOf` in `node.js` is two `indexOf` calls for exactly this reason |
| Rule 6 — every guard has a test that fails without it | 66 new cases; every refusal above has one |
| Terminal-escape / control-character injection (the `uf_pm::progress` row, pointed at a log) | Control characters are stripped and every string cut before a record is formatted |
| Credentials in diagnostics (the `.env` row's "no command prints a value") | Redaction by field name at every depth, and the flow never hands a token, a code or provider-supplied text to a logger |

The OAuth-specific ones, each with a case: login CSRF on a wrong `state`, a
replayed `state`, a callback in a browser that never began a flow, a subdomain
planting a cookie (`__Host-`), `SameSite=Lax` being required rather than
preferred, PKCE `S256` with no `plain`, open redirect on `?return=`, session
fixation on a second sign-in, the client secret in a header rather than a body,
and no token in any body or log line.

One decision worth flagging for review: **`secureCookies` defaults to `true`,
and the cookie name is resolved once for the deployment rather than per request
from the scheme.** Deriving it per request gives the flow two possible names,
and a reader that has to try both can be made to prefer an unprefixed one a
subdomain set — which is the fixation `__Host-` exists to prevent. Browsers
treat `http://localhost` as a secure context, so this costs nothing under
`uf dev`; the escape hatch is named for the one case that genuinely cannot be
secure.

## Temporal, not `Date`

@ubugeeei's review on `node.js`'s request timer, applied to every clock this
change reads. `@uniflowed/core/temporal` — the polyfilled Temporal
`@uniflowed/web`'s `Time` and `@uniflowed/hooks`' `useRenderedAt` already read —
is now a dependency of `@uniflowed/server`, and there is no `Date` left in what
this adds. It is not a spelling change: that module is uf's clock seam, so a
test or a host can install a clock and the values derived from it move with it.

| Was | Is |
| --- | --- |
| `const started = Date.now()` in three hosts, `Date.now() - started` for `durationMs` | `Temporal.Now.instant()`, and `elapsedMs(started)` — new in `@uniflowed/server/log` — which is `started.until(now).total({ unit: "millisecond" })` |
| `LogRecord.time: number`, `new Date(...).toISOString()` in both formats | `LogRecord.time: Temporal.Instant`, and each format builds its string from `toZonedDateTimeISO("UTC")` fields, always with three fractional digits: `Instant.toString()` drops the fraction on a whole second and `.` sorts below `Z`, so `…:05Z` would sort *after* `…:05.500Z` in a collector that sorts the timestamp string |
| `Session.expiresAt: number`, `TokenSet.expiresAt: number \| null`, `Date.now() + n * 1000` | `Temporal.Instant`, and `Temporal.Now.instant().add({ seconds })`. The `session` handler answers with the ISO string an instant serializes to, so a browser and a server component are told the same thing |
| a provider's `expires_in` added to a number, which cannot fail | bounded first — `NaN`, an infinity, a negative lifetime or one past a century reads as "the provider did not say" rather than throwing a `RangeError` out of the middle of a sign-in. `docs/security.md` has the row |
| a `Temporal.Instant` in a log field walked to `{}` | a value that knows its own text is logged as that text |

One number survives on purpose: `SessionStore.write`'s `expiresAt` is still
epoch milliseconds. That is the argument `@uniflowed/core/clock`'s header makes
for its own seam — a store's job is to cross into somebody's Redis, KV or table,
an epoch millisecond count is a primitive that serializes as itself and that
`PXAT` takes as it stands, and an `Instant` there would come back through a JSON
round trip as a string every adapter has to parse. Temporal on this side of the
seam, a number on the wire; `internal/oauth.js`'s header and the auth guide both
say so.

Nine cases hold it: a frozen clock deciding a record's timestamp, a `manualClock`
advanced 1500 ms deciding `durationMs`, the whole-second fraction, an instant in
a field, and five for the OAuth expiries.

## The `crates/` change

`uf_lib`'s registry mirrors each package's root exports and
`the_registry_names_exactly_what_each_package_exports` compares the two, so
`@uniflowed/server`'s entry in `crates/uf_lib/src/registry.rs` is red without
this. It is in the branch now — an earlier revision could not run cargo. The list
stays sorted:

```diff
             &[
                 "OutsideRequestError",
                 "after",
                 "cookies",
                 "draftMode",
                 "headers",
+                "logger",
+                "requestId",
             ],
```

Nothing else in `crates/` reads this package. The two new subpaths — `./log` and
`./oauth` — need no registry entry: no subpath appears in it today, and
`every_advertised_module_resolves_to_a_package` only checks registry → disk.

## Verification

Rebased onto `main` and run from a `cargo build --release --bin uf` of this
branch:

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — the only failures are the six in
  `uf_cli::vite` that bind a loopback socket, which this sandbox refuses
  (`Operation not permitted`); they pass in CI
- `cargo test -p uf_fmt` — including
  `shipped_sources_are_formatted_idempotently`, which is what was red
- `uf fmt --check` — clean
- `uf lint` — 0 errors, the same 14 pre-existing warnings, none in changed files
- `uf test tests/library` — **2115 passed, 0 failed, 1 skipped**

`uf check` is not reported: it has many pre-existing errors on an untouched tree
and its count depends on cache state (#564).

### What was red, and why

`Test` and `Upstream Flow Typecheck` were one failure with two names:
`uf_fmt`'s `shipped_sources_are_formatted_idempotently` found that under
`semicolons: false`, `node.js`'s `catch` block printed as
`;(options.log ?? processLogger()).error(…)` — a leading ASI guard standing
after a comment, which swallows the comment when the output is read back.
`crates/uf_fmt/src/flow/print/mod.rs` documents that shape as the one
`guard_above` cannot rescue, because the statement is the first in its block and
there is nothing above it to carry the `;`. The logger is resolved once into a
local instead, which is what `serve` a hundred lines below already does; there is
no parenthesized expression statement left to guard.

One existing test changed on purpose. `request-lifecycle.test.js`'s "settles a
request that failed" asserted the listener reported the failure to
`console.error` exactly once; it now reports two records — what went wrong, and
what the client was told — so the case asserts on those through a
`recordingLogger` instead of replacing a global.

## Worth follow-up issues

1. **`standalone.js` writes no access line.** Its handler answers an embedded
   asset and a prerendered document before it begins a request, so there is no
   single place with both the status and the context; making one means
   restructuring the streaming path, which deserves its own review. The other
   three hosts write it.
2. **`uf dev` and `uf preview` write no access line either.** Both answer inside
   `withRequest` in `@uniflowed/vite` rather than through `nodeListener`, and
   their terminal is rendered by the Rust side from an event channel — so the
   line wants to be an event rather than a write, which is a change to the
   protocol.
3. **A correlation id from a trusted proxy.** uf deliberately never reads
   `X-Request-Id`. A deployment that already assigns ids has a real need, and it
   is a second clearly-named field plus a statement of which hop is trusted.
4. **A durable session store.** The interface is the deliverable and
   `memorySessionStore` is honest about being one process's memory; an adapter
   shipping a Redis or KV implementation would make the seam load-bearing rather
   than only stated.
5. **`docs/security.md` rows that are still `todo`** and are now adjacent to
   real code: the RSC cache-poisoning row and the CSP-nonce row both touch
   what this change added `no-store` and `Vary` for.

Closes #505.
Closes #506.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/585"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791374865&installation_model_id=422833&pr_number=585&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F585&signature=3649ecb39f28d6facacb9bcd302dea7058d910a8f135ad0a9608193dc16ca3a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Applied the `registry.rs` edit — `logger` and `requestId` after `headers`; `cargo test -p uf_lib` 9 + 32 passed. Auto-merge armed.

Three things here are worth naming.

**`take` reading and removing in one step** is the design, exactly as you say. A read followed by a delete has a window and a window is all a replay needs — that is the difference between a `state` parameter that is single-use and one that is merely intended to be.

**Taking `redirect_uri` from the stored pending record rather than from the request** closes CVE-2025-29927's class at the one place OAuth is most exposed to it. Two requests with two `Host` headers cannot disagree, because only one of them wrote the record.

**The stdout hazard.** The Vite driver's stdout is a JSON control channel and `console.info` writes there on Node, so a logger that split by level would have corrupted it. Finding that before shipping, rather than after somebody's dev server started emitting unparseable frames, is the kind of thing that does not show up in a diff.

Changing  was right and correctly flagged: two records is the new truth, and asserting on them through `recordingLogger` is better evidence than counting `console.error` calls ever was.

I am wrapping up for the day, so your follow-ups are recorded here rather than filed — the access line missing from `standalone.js` and from `uf dev`/`uf preview`, a trusted-proxy correlation id, a durable session store, and the two still-`todo` rows in `security.md` that now sit next to real code.

packages/server/node.js
```diff
@@ -366,33 +330,65 @@ async function statFile(file: string) {
...
+     * not have to reach for a global to get it.
+     */
+    readonly log?: Logger,
   |},
 ): (incoming: NodeRequest, outgoing: NodeResponse) => Promise<void> {
   return async function listener(incoming: NodeRequest, outgoing: NodeResponse): Promise<void> {
     // Declared out here because `toRequest` is inside the `try`: a request that
     // could not even be built has no lifecycle to settle.
     let lifecycle: RequestLifecycle | null = null;
+    const started = Date.now();
```
ubugeeei: you must use Temporal (with polyfill)

ubugeeei: Done, and swept rather than patched.

Line 414 was `const started = Date.now()` — the start of a request's timer. It is now `Temporal.Now.instant()` from `@uniflowed/core/temporal`, which is the polyfilled Temporal `@uniflowed/web`'s `Time` and `@uniflowed/hooks`' `useRenderedAt` already read; `@uniflowed/server` now declares `@uniflowed/core`. Going through that module is not only spelling: it is uf's clock rather than the host's, so what a request line reports is something a test can install and move.

The same `Date` was in fourteen more places in this change, and all of them went:

- `node.js`, `edge.js`, `lambda.js` — `Date.now() - started` is now `elapsedMs(started)`, new in `@uniflowed/server/log`: `started.until(Temporal.Now.instant()).total({ unit: "millisecond" })`.
- `internal/log.js` — `LogRecord.time` is a `Temporal.Instant` rather than a number, stamped from `Temporal.Now.instant()`. Both formats lost their `new Date(...)` and build the string from `toZonedDateTimeISO("UTC")` fields, with three fractional digits always: `Instant.toString()` drops the fraction on a whole second and `.` sorts below `Z`, so `…:05Z` would land after `…:05.500Z` in any collector that sorts the timestamp string.
- `internal/log.js` — a field value that knows its own text (every Temporal type, and `Date`) is logged as that text. An `Instant` has no own enumerable properties, so an application logging the expiry it had just read out of `currentSession()` was logging `{}`.
- `oauth.js` — `Session.expiresAt` and `TokenSet.expiresAt` are instants, and every expiry is `Temporal.Now.instant().add({ seconds })` rather than `Date.now() + n * 1000`. The `session` handler answers with the ISO string an instant serializes to, so a browser and a server component are told the same thing spelled the same way.
- `oauth.js` — because Temporal's arithmetic has a range where a number silently had none, a provider's `expires_in` is now bounded before it reaches it: `NaN`, an infinity, a negative lifetime or one past a century reads as "the provider did not say" instead of throwing a `RangeError` out of the middle of a sign-in. `docs/security.md` has the row.
- `internal/oauth.js` — `memorySessionStore` holds expiries as instants and compares them with `Temporal.Instant.compare`.

One number survives, deliberately, and I would rather point at it than have you find it: `SessionStore.write`'s `expiresAt` is still epoch milliseconds. That is the argument `@uniflowed/core/clock`'s header makes for its own seam — a store's job is to cross into somebody's Redis, KV or table, an epoch millisecond count is an immutable primitive that serializes as itself and that `PXAT` takes as it stands, and an `Instant` there would arrive back through a JSON round trip as a string every adapter has to parse. Temporal on this side of the seam, a number on the wire; `internal/oauth.js`'s header and the auth guide both say so. Say the word if you would rather have no numbers anywhere and I will convert it.

Nine new cases hold it: a frozen clock deciding a record's timestamp, a `manualClock` advanced 1500 ms deciding `durationMs`, the whole-second fraction, an instant in a field, and five for the OAuth expiries.

Issue #388, opened 2026-09-06
rsc: decide whether a wrapped hook is client-only, instead of reporting that uf cannot


ubugeeei:
`uf build` and `uf dev` now say when they cannot answer: a called `use[A-Z]…`
that is not on `CLIENT_ONLY_APIS`, in a module the server runs, is reported as
`rsc/unclassified-hook-in-server` at severity `warn` — the first warning
`uf_rsc` has ever emitted, and deliberately not an error, because "uf does not
know" is not a contract violation. On uf's own documentation site that is four
call sites, three of them `useRoute()` and one `useTheme()`, which is #348's
worked example turned into output.

That closes the *silence*. It does not close the question, and this is the
issue for the question.

## What deciding it actually takes

Two halves, because a hook comes from two places, and only one of them is in
the module graph.

### 1. Hooks the project wrote — a fixpoint over the export graph

An exported function is a client hook when its body reaches a client-only API,
directly or through another export. `RscGraph` has the import edges and
`scan_exports` has the exported names; the two facts that are missing are both
per-module and both scannable without a parser:

* **which body a use site is in.** `ClientApiUse` records a line and column and
  no owner, so `useEffect` at 38:3 cannot be attributed to `useTheme`. The
  lexer already has `matching_open`/`matching_close`, so a declaration's body
  is the span from its `{` to that brace's match — enough to say which export
  contains a use site, and enough to be wrong only about code that redefines a
  binding mid-body.
* **which local name an import binds.** `ImportSpecifier` keeps the specifier
  string and drops the bindings, so `import { useRoute } from "…"` cannot be
  connected to the `useRoute()` two lines down. `imports_from_tokens` walks the
  clause already; keeping `{ imported, local }` pairs is a field, not a pass.

With those, classification is a monotone fixpoint over a finite set — mark an
export client-only when its body reaches a client-only API or calls a
client-only export, iterate until nothing changes — so it terminates and its
result does not depend on module order.

What stays unclassified, and must keep being reported rather than assumed
clean: `export *` re-exports, a hook reached through a computed property, and a
hook passed as a value and called somewhere else.

### 2. Hooks a package exports — the graph cannot see them at all

`analyze_project` skips `node_modules` (`IGNORED_DIRECTORIES`), and
`resolve_specifier` calls a bare specifier external. So `useRoute` from
`@uniflowed/router` — the exact example #348 leads with — has **no body in the
graph**, and the fixpoint above will never classify it. This is the half that
[...]

Issue #307, opened 2026-09-06
vite: every React Compiler diagnostic in a build says `a function` and names no file


ubugeeei:
A build of an application that uses `@uniflowed/form` prints this, twenty-odd times:

```
! [plugin uf:flow] a function: This value cannot be modified
! [plugin uf:flow] a function: Cannot access refs during render
! [plugin uf:flow] a function: [InferMutationAliasingEffects] Expected value kind to be initialized
```

No file, no line, no column, and no function name. There is nothing in the message a reader can act on, and nothing that says whether the code at fault is theirs.

## Reproduction

```sh
uf create app react repro
cd repro
```

Add one client component:

```js
"use client";
// @flow
import { useForm } from "@uniflowed/form";

export component Form() {
  const { register } = useForm({ defaultValues: { email: "" } });
  return <input {...register("email")} />;
}
```

render it from `app/_uf.page.js`, and build:

```
$ uf build 2>&1 | grep -c "plugin uf:flow"
2
$ uf build 2>&1 | grep "plugin uf:flow" | sort | uniq -c
   2 ! [plugin uf:flow] a function: [InferMutationAliasingEffects] Expected value kind to be initialized
```

A fuller app — the tutorial application being written for #304, which uses `@uniflowed/form`, `@uniflowed/state` and `@uniflowed/ui` — prints 40 of them, all "a function", across two Vite passes (client and server), which is also why every one appears twice.

## Where it comes from

`packages/vite/index.js:197-199`:

```js
for (const diagnostic of out.diagnostics) {
  this.warn?.(`${diagnostic.function ?? "a function"}: ${diagnostic.message}`);
}
```

The transform hook knows `id`; the warning does not carry it. Rollup's own `this.warn` normally attaches the module id when it is given a position, and a bare string gets neither. `diagnostic.function` being absent for every one of these suggests the React Compiler is reporting on an anonymous inner function, so the fallback string is doing all the work.

## What "good" would look like

* The module the diagnostic belongs to, in the message — a reader who cannot tell their own module from a dependency's cannot tell whether to act.
* A line and column where the compiler gives one, passed as `this.warn({ message, id, loc })` so Vite formats it like every other build warning.
* One report per diagnostic rather than one per Vite pass, or a note saying which pass it was.
[...]

PR #611, opened 2026-09-07
fix(transform, vite): a React Compiler finding that names its function and points at itself

### LINKED ISSUE #371: transform: `CompilerDiagnostic.function` can never be set, so no React Compiler finding can name a function
`crates/uf_transform/src/compiler.rs` reads a finding's function name out of the logger event:

```rust
let function = event
    .get("fnName")
    .and_then(Value::as_str)
    .map(str::to_owned);
```

No event that becomes a diagnostic has that field. `LoggerEvent` in `upstream/react/compiler/crates/react_compiler/src/entrypoint/compile_result.rs` declares `fnName` on exactly one variant, `CompileSuccess` — and `diagnostic()` returns `None` for `CompileSuccess`. `CompileError`, `CompileErrorWithLoc`, `CompileSkip`, `CompileUnexpectedThrow` and `PipelineError` carry `fnLoc` and no name at all.

So `CompilerDiagnostic.function` is `null` for every diagnostic uf will ever emit, structurally rather than by accident. It is not that the compiler happens to be reporting on anonymous inner functions, which is what ubugeeei-prod/uf#307 guessed from the outside; the name is simply not in the data.

That mattered more before #307: `packages/vite/index.js` printed `${diagnostic.function ?? "a function"}` and the fallback was doing all of the work, always. The Vite plugin now leads with the module and the position, so nothing is unreadable — but the field is a promise the type makes and the code cannot keep, and every host that reads it is reading a constant.

Two ways out, and they are not the same:

* Drop the field. Honest, and one less thing for a host to check. It also gives up on ever naming the function.
* Derive it. `fnLoc.start` is the position of the declaration, and `uf transform` already has the source it compiled. The identifier after `function`/`component`/`hook` at that position, or the name a `const X = () =>` binds, is the name a reader would use — and `app/Form.js:12:2: Cannot access refs during render (in Form)` is a better line than the same one without the parenthesis. It is a small amount of work over source text uf already holds.

The second is the one worth doing; the first is what the code is doing today without saying so.

Found while fixing #307.

### LINKED ISSUE #372: transform: `&& false` disables half of the diagnostic filter, and the half it disables is the interesting one
`crates/uf_transform/src/compiler.rs`:

```rust
fn diagnostic(event: &Value) -> Option<CompilerDiagnostic> {
    let kind = event.get("kind")?.as_str()?;
    if kind == "CompileSuccess" || kind == "PipelineError" && false {
        return None;
    }
```

`&&` binds tighter than `||`, so this reads `kind == "CompileSuccess" || (kind == "PipelineError" && false)` — the second clause is a constant `false` and the condition is `kind == "CompileSuccess"` with a decoration. `PipelineError` is reported like every other event.

Whoever wrote it meant to suppress `PipelineError` and then decided not to, or meant to try suppressing it and left the experiment in. Either way the source says something the compiler does not: a reader has to work out the precedence to learn that the line does nothing, and a `clippy::nonminimal_bool`-shaped lint does not fire on it because the comparison is not itself redundant.

What to do turns on a question the code does not answer, which is why this is an issue and not a patch: is a `PipelineError` — the compiler's own internal failure, carrying a `data` string rather than a user-facing reason — something an application author should be shown? If it is, delete the dead clause. If it is not, delete the `&& false` and let the filter do what it was written to do. Guessing is how the line got here.

Found while fixing ubugeeei-prod/uf#307, which touches the same function's output but not this.

ubugeeei:
A React Compiler finding used to arrive as `a function: This value cannot be
modified`. It now arrives as
`packages/hooks/dom.js:195:7: This value cannot be modified (in useLongPress)`.

What stood between those two lines was three reads of fields the compiler does
not put on these events — `fnName` for the name, `detail.reason` for the
message on the two kinds that carry `data`, and `detail.loc` for the position —
plus one clause that could not fire. #371 and #372 name the first and the last.
The `detail.loc` one is not in either issue; it only surfaced when the fixed
diagnostics were run over this repository's own packages, which is also where
the numbers below come from.

Closes #371
Closes #372

## The name (#371)

`CompilerDiagnostic.function` read `event["fnName"]`. `LoggerEvent` declares
`fnName` on exactly one variant — `CompileSuccess` — and `diagnostic()` returns
`None` for `CompileSuccess`. So `function` was `null` for every finding uf has
ever emitted, structurally rather than by accident, and `@uniflowed/vite`'s
`?? "a function"` fallback was doing all of the work every single time. #307
saw the fallback and inferred the compiler was reporting on anonymous inner
functions; it was not.

#371 offered two ways out — drop the field, or derive it from `fnLoc.start` and
the source. This takes the second, but off the tree rather than the text:
`fnLoc` is the function node's own `loc`, copied out of the AST uf handed in, so
indexing that tree by start position recovers the name with nothing to re-scan
and no guess about which declaration a position falls inside.

The naming rules are the compiler's own, so that a finding and a success about
the same function agree. Checked against `find_functions_to_compile` in
`react_compiler`:

* a `FunctionDeclaration` is its `id` (`fn_info_from_decl`);
* an arrow or function expression takes the name of the `const` it is the
  initialiser of, and only when the init is directly a function or a call
  (`enter_variable_declarator`);
* that name is carried through `memo` and `forwardRef`, bare or on `React`, and
  through no other callee (`get_callee_name_if_react_api`), so
  `const A = memo(() => …)` is `A` and `const A = wrap(() => …)` is anonymous;
* a function expression's own `id` is ignored exactly as the compiler ignores
  it, so `const A = function B() {}` is `A` — the binding the call site uses.

Anonymous stays anonymous. A name uf invented would send a reader looking for
text that is not in their file.

## The `&& false` (#372)

```rust
if kind == "CompileSuccess" || kind == "PipelineError" && false {
```

`&&` binds tighter than `||`, so the second clause was a constant and
`PipelineError` was already being reported. #372 asks which behaviour was meant
rather than which line to delete, so: **reporting stays, the dead clause goes.**

A `PipelineError` is the compiler failing on a function rather than the function
being wrong, and the author cannot fix the compiler — but the consequence is
theirs, because the function is left exactly as written and never memoised.
That is precisely what this module's `panicThreshold: "none"` promises will be
reported and not swallowed, and hiding the only event that says so would let a
`component` opt out of compilation with nothing in any log. `CompileSkip` and
`CompileUnexpectedThrow` are the same shape and are already reported; holding
back one of the three has no principle behind it. The noise a suppression would
have controlled is controlled where it belongs — the Vite plugin already holds
back a dependency's findings and counts them in one line.

What the objection was actually about is the message, and that is fixed here:
`PipelineError` and `CompileUnexpectedThrow` carry their text in `data`, not in
a `detail.reason`, so reading only the latter left both reporting the bare word
`PipelineError`.

## What the sweep found

Running the fixed diagnostics over `packages/`, `examples/`, `docs/` and
`tests/` — 355 modules, 173 findings:

**Every finding is named.** 74 distinct functions, zero anonymous. All 173 used
to print `a function`.

**Zero `PipelineError`.** Also zero `CompileSkip` and zero
`CompileUnexpectedThrow` — all 173 are `CompileError`. So #372's decision
changes nothing about what this repository reports; it matters for user code,
and the message fix is what matters here.

**A third field that is not there.** A finding's position came from `fnLoc`,
falling back to `detail.loc` — and this compiler never populates `detail.loc`
(`CompilerErrorDetailInfo::loc` is `skip_serializing_if` and is `None` on every
event uf has seen). The position is one level further down, on the
`detail.details` items. So the fallback written to make a finding precise was
the same kind of dead read as the `&& false`, and every finding reported the
line its `component` or `hook` keyword is on.

That was not cosmetic. `reportDiagnostics` de-duplicates on kind, position and
message, so findings that differed only in position were indistinguishable and
collapsed: **173 findings printed as 119 lines.** The 54 that vanished were not
repeats. `DatePickerInput` alone lost four:

| before | after |
| --- | --- |
| `date-picker.js:254:8: Cannot access refs during render` | `date-picker.js:256:34` — ref passed to a function |
| *(collapsed)* | `date-picker.js:262:31` |
| *(collapsed)* | `date-picker.js:265:33` |
| *(collapsed)* | `date-picker.js:268:34` |
| *(collapsed)* | `date-picker.js:277:24` — `.current` read in a ref callback |

Findings now point at the expression the compiler objected to, and name the
function around it, so the two halves of the line say different things. With
precise positions there are no duplicates left in this repository at all — the
comment claiming the compiler "reports it once per pass that noticed it, three
times for one `ref`" was reading this artefact, and is corrected rather than the
de-duplication removed (it still guards a genuine repeat of one site).

## Nothing in `packages/` is a bug

101 of the 173 are in `packages/`, and none of them is a defect to fix:

* **78 "This value cannot be modified" / 17 "Cannot access refs during render"** —
  refs written from event handlers, effects and ref callbacks, and refs carried
  through `composeRefs`/`composeHandlers`/`withoutComposed`, which the compiler
  cannot see through. `useLongPress` already carries the comment saying so
  ("Written and read only from event handlers and the effect's cleanup, never
  during a render"), and `useStableCallback` is React's own `useEvent` RFC
  shape, which is not memoisable by construction.
* **61 "Cannot reassign variables declared outside"** — every one in
  `tests/library/*.test.js`, where a test component assigns to a module-scope
  `let` so the assertion can read it. A test idiom, not shipped code.

These are the compiler declining to memoise deliberate escape hatches. It leaves
them as written and reports them, which is what this module is for, so nothing
is suppressed and no source line is annotated to quiet a build.

## Two implementations, reconciled

This branch grew two answers to #371 in parallel, and `4046a0a` merges them into
one. Both read the name off `fnLoc`, because that is all a finding carries; they
differ in what they read it out of.

The first, `declared_name`, scanned the source text forward from the position
and matched `function`/`component`/`hook`/`const` at it. That is right for a
declaration and cannot work for an arrow: `fnLoc` on an
`ArrowFunctionExpression` is the arrow's own `loc`, so it points at the `(` of
`() => {}` and never at the `const` to its left. `const Card = () => …` and
`const Card = memo(() => …)` — the shape #371's second bullet asks for by name —
came back anonymous. Its unit tests passed because they call the helper with the
declaration's position, which the pipeline does not produce for an arrow.

The AST index kept here gets the same answers for declarations, gets the arrow
and the `memo`/`forwardRef` wrapper as well, and is checked against the
compiler's own `find_functions_to_compile`. Both sides reached the same
conclusion on #372 and both fixed the `data`-carried message, so that half
merged without argument. The earlier commit is preserved in the history.

## Not included: #388

#388 wants `uf_rsc` to decide whether a wrapped hook is client-only, which needs
a fixpoint over the export graph plus a `client_only` set in
`crates/uf_lib/src/registry.rs`. It shares no file with this change — `uf_rsc`
does not reference `uf_transform` at all — and #388 says itself that the React
Compiler route answers a different question and "would not have caught
`useRoute`". Left open.

## Tests

Seven new tests in `compiler.rs`, checked to fail against the code they
describe and pass after it: declaration / `hook` / arrow / `memo`-vs-other-call
naming, `data`-carried messages on both kinds that use them,
only-a-success-is-held-back, and the expression-not-declaration position with
its no-located-item fallback.

One existing test changed. `a_react_compiler_finding_names_its_file_and_line`
in `crates/uf_cli/tests/vite_styles.rs` asserted `mine.len() == 1` for a fixture
whose component touches `box.current` three times — an assignment, the read on
the right of it, and a read in the JSX. That one line was the collapse, not the
truth, so the test now asserts what its name and its message always meant:
three accesses give three positions, each names `Home`, and no position repeats
across the two Vite passes. Strictly more than it checked before.


ubugeeei:
This pull request grew a second implementation of #371 in parallel with `01abc06`, and `4046a0a` merges the two rather than replacing one branch with the other — the original commit is still in the history, and the push was a fast-forward.

**What changed, and why.** `declared_name` scanned the source text forward from the finding's position for `function`/`component`/`hook`/`const`. That is correct for a declaration. It cannot reach an arrow, because `fnLoc` on an `ArrowFunctionExpression` is the arrow's own `loc` — it points at the `(` of `() => {}`, never at the `const` to its left — so `const Card = () => …` and `const Card = memo(() => …)` came back anonymous. Those are the shape #371's second bullet asks for by name. The helper's unit tests pass because they call it with the declaration's position, which the pipeline does not produce for an arrow.

Indexing the Babel AST by each function's start position gets the same answers for declarations, gets the arrow and the `memo`/`forwardRef` wrapper too, and is checked line by line against the compiler's own `find_functions_to_compile`, so a finding and a `CompileSuccess` about one function agree on its name.

**What was kept.** Both sides reached the same conclusion on #372 — report `PipelineError`, delete the dead clause — and both found that `PipelineError` and `CompileUnexpectedThrow` carry their text in `data`. That half merged without argument.

**What is new on top.** Running the fixed diagnostics over this repository's own packages found a third read of a field the compiler does not populate: a finding's position fell back to `detail.loc`, which is always absent, so every finding reported the line its `component` or `hook` keyword is on. The plugin de-duplicates on kind, position and message, so that collapsed 173 findings into 119 lines — and the 54 it dropped were distinct sites, five of them different reads of a ref in `DatePickerInput`. The body has the numbers.

Happy to split the position fix into its own pull request if that reads better.

Issue #425, opened 2026-09-06
cli: the tab-completion list of explainable commands is a hand-maintained subset of the real one


ubugeeei:
`uf explain <TAB>` offers the commands in `completion::EXPLAINABLE`.
`uf explain <name>` accepts the commands in `explain::KNOWN`. They are two
hand-maintained lists and nothing compares them.

They had already drifted: `EXPLAINABLE` was missing `install`, `upgrade`, `run`,
`exec` and `env`, among others — every one of which `uf explain` answers. So the
completion offers a subset, and a person who uses tab completion to find out
what is explainable is told less than the truth.

#287 added its four new commands to both lists by hand, which is the same
maintenance that produced the drift.

## What it needs

One list, or a test that fails when they differ. The test is the smaller change
and catches the case where a new command is added to `KNOWN` alone, which is the
direction the drift has always gone.

PR #658, opened 2026-09-08
ci(semver): the crates that cannot be compared are computed, not listed

ubugeeei:
**`Cargo Semver Checks` is currently failing on every open pull request.** It
released #637 and #638 from the merge queue within three minutes of each other,
and it will do the same to the rest.

## Why

The baseline is built from a copy of each crate *outside* the workspace, where
a relative path dependency no longer resolves. So a crate that reaches
`uf_flow` or `uf_check` — and through them the `upstream/flow` submodule — can
never be compared, and `semver-exclude.sh` exists to say so. It carried that
closure as a list somebody adds to:

```sh
submodule_path='uf_check uf_cli uf_doc uf_flow uf_fmt uf_lint uf_transform'
```

`uf_i18n` (#633) reaches `uf_flow`. It was new, so the script's *other* half —
the computed one — excluded it and the job stayed green. When the baseline
caught up it rejoined a gate it can never pass:

```
required by package `uf_flow v0.0.0-alpha.14 (…/semver-checks/target/…)`
    ... which satisfies path dependency `uf_flow` of package `uf_i18n`
```

`uf_lib` and `uf_router` reach `uf_flow` too and were never in the list either.
The run stops at the first crate it cannot resolve and `uf_i18n` sorts before
both, so they were latent behind it.

## The fix

The closure is read from the manifests, seeded with the two crates that name
the submodule directly, and iterated to a fixed point. On this tree:

```
uf_check uf_cli uf_doc uf_flow uf_fmt uf_i18n uf_lib uf_lint uf_router uf_transform
```

— the old seven plus exactly the three that were missing. Nothing that does not
reach `uf_flow` is excluded, so no coverage is given up.

A hand-maintained list of what to check is the same shape as the bug **#590**
fixed for the CI gate's `needs:` and **#425** for `uf explain`. This is the
third; the pattern is worth naming.

## The test

`tools/ci/test-semver-exclude.sh`, six cases against scratch workspaces with a
real git history, wired into `uf run ci` and the `Metadata` job:

- the seed and its dependents, and nothing else;
- the closure is **transitive**, not one level (`uf_router` → `uf_fmt` → `uf_flow`);
- a crate absent at the baseline is excluded;
- **and stays excluded once the baseline catches up, because it reaches `uf_flow`** — the case that broke the job;
- a new crate that reaches nothing is excluded now and **rejoins by itself** when the baseline has it.

## Verification

```
uf run semver:exclude:test   exit 0   (6 cases)
uf run ci:gate               exit 0
uf run scripts:parse         exit 0
uf fmt --check               exit 0
semver-exclude.sh HEAD~1     the ten above
```

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/658"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791435659&installation_model_id=422833&pr_number=658&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F658&signature=c9a7052048c387c755924d48fae141754019c5031c16bb7035f286149b5ef105"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -416,6 +416,7 @@ jobs:
       - run: ./target/release/uf run ci:gate:test
       - run: ./target/release/uf run ci:runtimes
       - run: ./target/release/uf run ci:runtimes:test
+      - run: ./target/release/uf run semver:exclude:test
 
   upstream-integrations:
     name: Upstream Integrations
--- a/tools/ci/semver-exclude.sh
+++ b/tools/ci/semver-exclude.sh
@@ -7,10 +7,21 @@
 # It names the `upstream/flow` submodule through a path dependency. The
 # baseline is built from a copy of the crate outside the workspace, where a
 # relative path no longer resolves — which is not something a crate can fix,
-# since a path dependency is relative by definition. The list below is the
-# closure of the crates reaching `uf_flow` or `uf_check`, whether or not the
-# feature using it is enabled, and it grows as more crates print from or read
-# the parser's syntax tree. See docs/architecture.md.
+# since a path dependency is relative by definition. That is the closure of
+# the crates reaching `uf_flow` or `uf_check`, whether or not the feature
+# using it is enabled. See docs/architecture.md.
+#
+# The closure is **computed**, and it used to be a list somebody added to.
+# That list went stale the moment `uf_i18n` was added: it reaches `uf_flow`,
+# it was new so the second half below excluded it, and the release after that
+# the baseline caught up and it rejoined a gate it can never pass. The job
+# then failed on every pull request, at `uf_i18n` — which is alphabetically
+# first among the crates the list was missing, so `uf_lib` and `uf_router`
+# were latent behind it and would have been next.
+#
+# A hand-maintained list of what to check is the same shape as the bug
+# ubugeeei-prod/uf#590 fixed for the CI gate and ubugeeei-prod/uf#425 for
+# `uf explain`. This reads the manifests instead.
 #
 # Or it did not exist at the baseline revision. A new crate has nothing to be
 # compared against, and cargo-semver-checks ends the whole run rather than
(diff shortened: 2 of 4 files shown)


ubugeeei:
Superseded by #657, which landed a few minutes ago and does the same thing: the closure of crates reaching `uf_flow`/`uf_check` is computed from the manifests instead of written down, with `tools/ci/test-semver-exclude.sh` and a `ci:semver:test` task behind it. We arrived at this independently and at the same time — apologies for the collision.

Closing this one rather than #657 for two reasons, and if either is wrong the fix is small enough to port:

- **`[dev-dependencies]` are excluded from the edges there.** `uf_lib` and `uf_router` reach `uf_flow` only through a dev-dependency; `cargo update` resolves the baseline fine without those, so counting them would have quietly taken two crates *out* of the semver gate. That is the direction of this change with no red to warn anybody, so it is the one worth pinning — `#657`'s third case is exactly it, and removing the section walk fails it.
- **A cycle case.** The closure is a fixed point, and one that does not terminate is a job that hangs rather than fails.

If there is anything here that #657 missed, please say and I will port it.

ubugeeei:
Closing this: **#657 landed the same fix and is the better version of it.**

I wrote this after `Cargo Semver Checks` released #637, #638 and #644 from the
merge queue in the space of five minutes, without seeing that #657 was already
in flight for the same failure. Both compute the closure instead of listing it,
and both catch `uf_i18n`. The difference is what else they catch:

| | this branch | #657 |
| --- | --- | --- |
| closure | ten crates | eight |
| extra | `uf_lib`, `uf_router` | — |

**#657 is right and this is wrong.** `uf_lib` reaches `uf_flow` and
`uf_router` reaches `uf_fmt` **only through `[dev-dependencies]`**, and the
baseline builds fine with those — so excluding them takes two crates out of
the gate for nothing, which is the opposite of what the file is for. #657 walks
the manifest by section and skips `[dev-dependencies]`; this one greps, cannot
tell the sections apart, and over-excludes.

Its own comment says so, at the top of the edge extraction:

> `[dev-dependencies]` are skipped, and that is the whole reason this walks
> sections rather than grepping the file.

Nothing here is worth salvaging on top of it — the test cases overlap, and
#657 has its own. Closed unmerged.

PR #666, opened 2026-09-08
fix(infra, security): the documentation site sends a content security policy

### LINKED ISSUE #598: infra: the docs worker sends no content-security-policy, on the site that argues about CSP
`infra/cloudflare/workers/docs.js` sets `x-content-type-options`,
`referrer-policy` and `permissions-policy`, and **no
`content-security-policy`**.

Found while building the security scan for #524, which deliberately did not
cover response headers: the only headers in this repository are set by the
Cloudflare workers, which are a *deployment* rather than something `uf build`
writes, and asserting a server's behaviour by grepping its source would be
asserting a file rather than a behaviour.

So this is a change to a worker, and the check that would hold it is a
different one from `security:scan`.

## Why it matters here specifically

The documentation site is where uf argues about security. `docs/security.md`
has a row for "XSS via CSP nonce handling and `beforeInteractive` scripts", and
the site making that argument serves no CSP at all.

## Expected

- The docs worker sends a `content-security-policy` that is actually derived
  from what the site loads, rather than a copied-in default.
- Something fails when it stops being sent, or when the site starts loading
  something the policy forbids — a header assertion against the deployed
  worker, or a test of the worker's own handler.

## Worth deciding at the same time

`docs/dist/docs/.vite/manifest.json` — Vite's module-to-chunk map — ships with
the site. Nothing reads it to answer a request, so it is disclosure with no
reader rather than a vulnerability, which is the same argument
`docs/security.md` already makes about `uf-rsc-manifest.json`. Turning it off
is a build-configuration decision; it is noted here rather than made a rule.

ubugeeei:
`infra/cloudflare/workers/docs.js` set `x-content-type-options`, `referrer-policy` and `permissions-policy`, and **no `content-security-policy`** — on the site where uf argues about security, whose own `docs/security.md` has a row about CSP.

## Derived, not copied

Every directive was read off the build. The site serves same-origin scripts, stylesheets and images and nothing else:

```
external origins loaded   none  (every https:// in the HTML is a link in prose)
@font-face / url() in CSS none
data: URIs                none
cross-origin fetch        none
<iframe>/<object>/<embed> none
inline style attributes   none
```

So every fetch directive is `'self'`, and the ones for things the site never does are `'none'`:

```
default-src 'self'; script-src 'self' 'sha256-…'; style-src 'self'; img-src 'self';
font-src 'self'; connect-src 'self'; manifest-src 'self'; object-src 'none';
frame-src 'none'; frame-ancestors 'none'; base-uri 'self'; form-action 'self'
```

Three worth naming:

- **`style-src` needs no `'unsafe-inline'`.** That is the commonest reason a policy is weakened, and this build emits no inline `style` attribute and no `<style>` element — a fact about the site, asserted, rather than an aspiration.
- **`base-uri 'self'`**, without which one injected `<base>` redirects every relative script URL in the document to another origin, and `script-src 'self'` does not stop it.
- **`frame-ancestors 'none'`**, so a framed copy of the manual fails.

## The one inline script

The theme bootstrap has to be inline and synchronous — a stored dark preference applied after the first frame is a white flash, and no CSS can express "read localStorage" — so it is allowed **by hash** rather than by weakening `script-src` to `'unsafe-inline'`.

`<script type="application/ld+json">` needs no hash of its own: it is a data block the parser never executes. That is as well, because its content differs per page.

## The test is the hash

The hash is the part that can drift, so it is the part that fails. `tests/library/docs-csp.test.js` computes it from `themeBootstrap`'s exported string and compares it with the worker's constant:

```
$ # after changing STORAGE_KEY in docs/app/_design/theme.js
✗ the inline theme bootstrap is allowed by hash > hashes to what the worker's policy names
```

That is the difference between a caught edit and a theme that silently stops working behind a policy nobody re-read.

The rest asserts the shape rather than the string — no `unsafe-inline` or `unsafe-eval` anywhere, `'none'` where it belongs, and **no directive naming any origin at all** — and checks the headers on an asset response as well as an HTML one, because a stylesheet served without `x-content-type-options` is the sniffing case and a policy that only reached HTML would be a policy on one response in ten.

`docs/security.md` gains the row, naming the test.

Ten tests, green.

Closes #598

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/666"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791445454&installation_model_id=422833&pr_number=666&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F666&signature=9acedc2915515f8a7816101910d12e4e1e370b5d8672e9634349d5cc404dba39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Superseded by #652, which closes #598 as part of a wider change and does it better on two counts I got wrong:

- **It hashes all four inline scripts a build emits, including the `application/ld+json` block.** Mine hashed only the theme bootstrap, on the reasoning that a data block is never executed and so is not subject to `script-src`. #652 says browsers do subject it, and hashing it costs nothing while getting that wrong breaks the site's structured data behind a policy nobody would re-read. The safe direction is the one that hashes it.
- **`tools/ci/docs-csp.sh` re-derives the policy from `docs/dist/docs` on every `uf run docs:verify`.** Mine held the hash to the exported string, which catches an edit to the bootstrap and nothing else — it would not have noticed the site starting to load a font, which is the failure the issue actually names.

Closing rather than merging the two: everything mine asserted is either already there or weaker.

Sorry for the collision — I should have checked the open pull requests for `Closes #598` before starting, and I have gone through the rest of the queue now so it does not happen again.

Issue #668, opened 2026-09-08
perf: `uf lint` allocates 840,000 times and 72 MiB for one 111 KiB module, and 94% of it is one rule building a Babel AST


ubugeeei:
Measured with a counting allocator over `packages/router/internal/runtime.js` (111,272 bytes), release build, five runs after a warm-up:

```
time / run     88.40 ms
allocs / run   839,171
bytes / run    71.78 MiB
```

That is **7.5 allocations per source byte** and **660× the input in bytes allocated**. For comparison, `uf fmt` over the same file is 5.3 ms, 36,211 allocations and 7.78 MiB — the linter costs sixteen times the memory and seventeen times the time.

## Where it goes

Per-phase, same file and run count:

| phase | allocs/run | MiB/run |
| --- | ---: | ---: |
| `run_react_tree_rules` | **789,874** | **64.77** |
| `run_tree_rules` | 23,703 | 3.20 |
| `run_flow_syntax` | 23,692 | 3.20 |
| `run_flow_unclear_type` | 1,871 | 0.01 |
| `run_react_compiler_rules` | 19 | 0.46 |
| `FileScan::new` | 2 | 0.15 |
| `mask_inline_comments` | 0 | 0.00 |
| everything else | ≤4 | ~0 |

**One rule is 94% of the allocations and 90% of the bytes.** Every other rule together is under 6%.

## Why that rule

`run_react_tree_rules` calls `uf_transform::babel_ast`, which is 789,496 of those 789,874 allocations — the rule's own work is 378. Splitting it:

| step | allocs | MiB |
| --- | ---: | ---: |
| `estree::parse` | 247,542 | 27.16 |
| `lower::lower` | 232,386 | 12.66 |
| `babel::to_babel` | 309,568 | 24.90 |

**There is no hot spot.** All three are the same order, because all three work on `serde_json::Value` — a `String` and a `Map` per node, three times over. This is the shape of the pipeline rather than a defect in any part of it, which is why it wants a decision rather than a patch.

## What it is for

Two rules: `react/derived-state` and `react/redundant-memo`. The gate is textual — the file must mention `useEffect`, or `useMemo`/`useCallback` — and `runtime.js` mentions both `useEffect` and `useState` three times each, so the work is legitimately wanted here. This is not a case of a rule running when it cannot fire.

## Options, none of them free

[...]

PR #673, opened 2026-09-08
perf(transform): 31% fewer allocations building the Babel tree

ubugeeei:
Refs #668.

`uf lint` allocates 830,581 times and 67.79 MiB for one 111 KiB module, and 94% of it is `uf_transform::babel_ast`. #668 lays out four options and calls the pipeline's `serde_json::Value` shape the real answer. **This is none of those four** — it is the waste that has nothing to do with the shape, found by reading the three walks the tree goes through.

| | before | after | |
| --- | ---: | ---: | ---: |
| allocations / run | 830,581 | 575,198 | **−30.7%** |
| bytes / run | 67.79 MiB | 56.04 MiB | **−17.3%** |

Measured on `packages/router/internal/runtime.js` (113,734 bytes), release build, five runs after a warm-up.

## Three changes, all the same mistake

**The key list — the largest.** `transform_post_at` and `finalize` both collected every key of every node into a `Vec<String>` before walking its children, because `get_mut` cannot run while `keys` is borrowed:

```rust
let keys: Vec<String> = object.keys().filter(…).cloned().collect();
for key in keys { match object.get_mut(&key) { … } }
```

Neither loop adds or removes a key — both only write into one — so `iter_mut` is the same walk in the same order, with no vector and no copies. `transform_post` carries both the lowering *and* the Babel conversion, so this one was paid twice a module.

**The type name.** `convert`, `finalize` and `strip_node` each read the node's type as a `&str` and immediately `to_owned` it, because the arms below need the node mutably — one `String` per node per pass, for a name compared once and dropped. Classified into a `Copy` enum while it is still borrowed instead, which costs nothing and reads better at the match.

**The `loc` object.** `finalize` replaced every node's `loc` with a fresh `json!({ start, end })` — three maps and six keys, per node. The translator has already attached one of exactly that shape (`estree::parse` asks for `include_locs`); only its *numbers* are wrong, because its columns count code points where everything downstream counts UTF-16 units. Overwriting four numbers into slots that already exist allocates nothing.

## What this changes about the output

Nothing about the tree, and one thing about how it serializes — and I checked rather than assumed.

`babel_ast` was run over **all 277 modules under `packages/`**, before and after, and the trees compared as parsed JSON:

```
semantically identical: 277
semantically DIFFERENT: 0
```

Compared as *text*, all 277 differ, in exactly one way:

```
before: "loc":{"start":{"line":76,"column":0},"end":{"line":721,"column":1}}
after : "loc":{"end":{"line":721,"column":1},"start":{"line":76,"column":0}}
```

`loc`'s key order is now the translator's rather than the `json!`'s, and `serde_json` is built with `preserve_order`, so that is observable in a serialized tree.

Nothing in this repository reads a Babel tree by key order — the React Compiler crate deserializes it, and the printer, the scope analysis, the memo rule and `uf_lint`'s tree rules all address it by name — so this is a difference rather than a change in behaviour. Flagging it because it is the one thing a reviewer should decide about rather than take on trust: if you would rather the bytes stayed identical, the `loc` change is separable and worth 78k of the 255k saved.

## The reproducer is checked in

`examples/alloc_report.rs` is how the figures above were taken, and its `--dump` is how the trees were compared:

```bash
cargo run --release --example alloc_report -p uf_transform -- packages/router/internal/runtime.js
```

#668's numbers were otherwise a claim rather than something the next person could re-take, and this pipeline is going to be touched again.

## Not closing #668

The issue's real question — whether the two `react/*` rules should read the native Flow AST instead, or whether the Babel tree should stop being `serde_json::Value` — is untouched. 575,198 allocations for one module is still 5 per source byte.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/673"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791454625&installation_model_id=422833&pr_number=673&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F673&signature=c715d96ef3cb3329fce2fcdceacf47fed5b130391bd1b43c9cb3585a6bd63e62"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3999,6 +3999,7 @@ dependencies = [
  "uf_check",
  "uf_flow",
  "uf_infra",
+ "uf_profiler",
 ]
 
 [[package]]
--- a/crates/uf_transform/Cargo.toml
+++ b/crates/uf_transform/Cargo.toml
@@ -37,6 +37,11 @@ similar-asserts.workspace = true
 # only one that can see every entry point.
 uf_check = { path = "../uf_check" }
 
+# For `examples/alloc_report.rs`: the counting allocator that produced the
+# figures in ubugeeei-prod/uf#668, so the next person to touch this pipeline can
+# reproduce them rather than take them on trust.
+uf_profiler = { path = "../uf_profiler" }
+
 [[bench]]
 name = "transform"
 harness = false
--- a/crates/uf_transform/examples/alloc_report.rs
+++ b/crates/uf_transform/examples/alloc_report.rs
@@ -1,0 +1,56 @@
+//! What `babel_ast` costs, in allocations and bytes, for ubugeeei-prod/uf#668.
+//!
+//! Not a test and not a bench: a number to put in a commit message, and the
+//! thing to re-run after a change to see whether it moved. Run it as
+//! `cargo run --release --example alloc_report -p uf_transform -- <file>`.
+
+use uf_profiler::{AllocSnapshot, CountingAllocator, Window};
+
+#[global_allocator]
+static GLOBAL: CountingAllocator = CountingAllocator::new();
+
+fn main() {
+    let path = std::env::args()
+        .nth(1)
+        .unwrap_or_else(|| String::from("packages/router/internal/runtime.js"));
+    let source = std::fs::read_to_string(&path).expect("read the module");
+
+    // `--dump`: the tree itself, so a change meant only to cost less can be
+    // held against the tree it used to produce. `serde_json` is built with
+    // `preserve_order`, so this is sensitive to key order as well as to
+    // content — which is how the one difference this change does make was
+    // found rather than assumed away.
+    if std::env::args().any(|arg| arg == "--dump") {
+        let (file, _) = uf_transform::babel_ast(&source).expect("parses");
+        println!("{}", serde_json::to_string(&file).expect("serializes"));
+        return;
+    }
+
+    println!("{path}: {} bytes", source.len());
+
+    // Warm up: the first run pays for whatever the parser initialises once.
+    let _ = uf_transform::babel_ast(&source).expect("parses");
+
+    CountingAllocator::enable();
+    let window = Window::open();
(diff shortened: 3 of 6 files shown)


ubugeeei:
Following up on the one thing I flagged for a decision — the `loc` key-order change — with the check rather than the argument.

I went looking for anything that turns a Babel node back into text, since that is the only way the order could escape. There are two `serde_json::to_string` calls in the pipeline and neither takes a node:

- `print.rs:935` — `fn string(&mut self, value: &str)`, which quotes a JavaScript string literal.
- `lower/builders.rs:54` — the same, building a literal's `raw` text.

Nothing hashes an AST either, so there is no build cache keyed on the serialization that this would invalidate or split.

That leaves the consumers I listed, all of which address the tree by name: the React Compiler crate deserializes it into its own types, and `print.rs`, `scope.rs`, `memo.rs` and `uf_lint`'s `react_tree.rs` all read fields through `get`. So the order is unobservable outside a deliberate `to_string` of the whole tree — which is what `examples/alloc_report.rs --dump` does, and is how I found it.

Still happy to drop the `loc` half if you would rather the bytes stayed identical; it is worth 78k of the 255k and is one hunk.

PR #680, opened 2026-09-08
fix(host, transform): uf's own build directory is not Flow source

### LINKED ISSUE #679: build: `uf build --compile` stops working when a site's assets pass 8 MB, because uf parses its own generated bundle as Flow
`uf build --compile` packs the whole output directory into one JavaScript module so a standalone binary has everything inside it:

```js
// Generated by `uf build --compile` from dist. Not checked in, not edited.
export const assets = JSON.parse("{\"index.html\":{\"type\":\"text/html…\",\"body\":\"PCFkb2N0eXBl…\"}}");
```

Every asset, base64'd, in one string literal. It grows with the site. Past eight megabytes the build fails:

```
✗ Build failed with 1 error:

  [plugin uf:flow] docs/.uf/build/compile/assets.js
  TransformError: source is 8442536 bytes, over the 8388608 byte ceiling
```

`MAX_SOURCE_BYTES` is 8 MiB, and its comment says why it exists: *"A dependency can put a generated or hostile file in `node_modules`; every scan in uf has an explicit ceiling rather than trusting its input."*

That reasoning is right and does not apply here. This file is not input — **uf wrote it, a moment earlier, out of files uf had just written**. The ceiling that exists to distrust a dependency is being applied to uf's own output, and it caps how large a site `--compile` can carry.

## Why it happens

`isFlowModule` claims every `.js` outside `node_modules`, and `.uf/` is outside `node_modules`. So uf hands its own generated bundle back to itself to parse as Flow.

Nothing under `.uf/` is Flow. It is uf's working directory: the bundled server, the deploy adapters, the asset module — all of it already-plain JavaScript that uf emitted. Parsing it was waste at every size before it became a failure at one:

```
docs/.uf/build/compile/server.js    13,234,886 bytes
docs/.uf/build/compile/assets.js     8,342,956 bytes
```

`server.js` is larger still and does not fail only because it is the bundler's *output* rather than an input the plugin is asked about.

## How it surfaced

It failed `compile_writes_one_file_that_serves_the_site_from_an_empty_directory` on #643, and would fail on any branch once the docs site crosses the line — it is currently within about fifty kilobytes of it, so this is not a #643 problem. Anyone whose site has more assets than uf's own docs already cannot use `--compile`.

## Fix

`.uf/` is not Flow, for the same reason `node_modules` is not: nothing in it was written by hand. That is one rule expressed in three places which a contract test already holds together — `isFlowModule`, `FLOW_MODULE_PATTERN` and `uf_transform::is_flow_module` — so all three learn it at once, with rows in `tests/library/flow-modules.test.js` covering the nesting cases and the near-misses (`/p/.uf-notes/x.js`, `/p/my.uf/x.js`, `/p/app.uf.js` are all still source).

Raising the ceiling would be the wrong fix: the number is not too small, the file was never input.

ubugeeei:
Closes #679. **This unblocks #643**, which is currently red on it and has been holding the merge train.

`uf build --compile` packs the output directory into one JavaScript module so a standalone binary carries everything inside it — every asset, base64'd, in a single string literal. It grows with the site, and past eight megabytes the build stops:

```
✗ Build failed with 1 error:
  [plugin uf:flow] docs/.uf/build/compile/assets.js
  TransformError: source is 8442536 bytes, over the 8388608 byte ceiling
```

`MAX_SOURCE_BYTES` says why it exists: *"A dependency can put a generated or hostile file in `node_modules`; every scan in uf has an explicit ceiling rather than trusting its input."* That is right, and it is not what happened. **uf wrote this file, a moment earlier, out of files uf had just written.** A ceiling meant to distrust a dependency was deciding how large a site `--compile` can carry.

## The rule

`isFlowModule` claims every `.js` outside `node_modules`, and `.uf/` is outside `node_modules` — so uf handed its own bundle back to itself to parse as Flow. Nothing under `.uf/` is Flow: the bundled server, the deploy adapters and the asset module are all output uf emitted as plain JavaScript. Parsing them was waste at every size before it was a failure at one:

```
docs/.uf/build/compile/server.js   13,234,886 bytes
docs/.uf/build/compile/assets.js    8,382,332 bytes
```

`server.js` is larger and escapes only because it is the bundler's *output* rather than an input the plugin is asked about.

So: `.uf/` is not Flow for the same reason `node_modules` is not — nothing in it was written by hand.

## Three expressions, one rule, already under contract

`packages/host/transform.js` documents that `FLOW_MODULE_PATTERN.test(path) === isFlowModule(path)` exactly, and `uf_transform::is_flow_module` mirrors both. All three learn this together, and each is load-bearing:

| removed | fails |
| --- | --- |
| `isFlowModule`'s check | `flow-modules.test.js` |
| `FLOW_MODULE_PATTERN`'s lookahead | `flow-modules.test.js` |
| the Rust mirror | `ufs_own_build_directory_is_not_flow` |

The near-misses are in the table because a substring test would get them wrong — `/p/.uf-notes/x.js`, `/p/my.uf/x.js` and `/p/app.uf.js` are all still source, and `.uf/` nested inside a package is still rejected for the earlier reason.

## What I could and could not verify locally

`compile_writes_one_file_that_serves_the_site_from_an_empty_directory` passes here — but it passes **with or without this change**, because my `assets.js` is 8,382,332 bytes: **6,276 bytes under the ceiling**. I cannot reproduce CI's failure on this machine, and I am not going to claim I did.

What is proven is the rule itself, by the contract tests above, and the causal chain: the plugin calls `transform` only when `isFlowModule(id)` is true, so making it false for this path means `TransformError` cannot be raised for it. CI on #643 is what will show it end to end.

That the docs site sits six kilobytes under the line is also the point: this is not a #643 problem. Any branch that adds a few pages tips it over, and any user whose site has more assets than uf's own docs already cannot use `--compile`.

Raising the ceiling would have been the wrong fix. The number is not too small; the file was never input.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/680"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791462117&installation_model_id=422833&pr_number=680&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F680&signature=34c83498c911862f49eda4b86dd47b6005b781877964a6f181a51fb3f1d42f1b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
CI caught what my local runs did not, and it was a real conflict rather than a flake.

Excluding `.uf/` from the Flow transform also excluded **every test fixture**. `Project::new` builds them at `<repo>/.uf/test-projects/`, so eight asset tests failed with `Flow is not supported` pointing at their own `app/_uf.page.js`:

```
[PARSE_ERROR] Flow is not supported
   ╭─[ app/_uf.page.js:1:1 ]
```

I had checked the generated JS under `.uf/` — the bundled server, the deploy adapters, the asset module — and none of it is Flow. What I had not checked is that the *test harness* puts project source in there too, which the fixtures' own doc comment explains: they need `node_modules` above them, and a system temp directory has none.

The rule is right and the location is wrong. `.uf/` is uf's output; source living inside a build directory is what made the two collide. The fixtures move to `.uf-test-projects/` — a sibling rather than a child, keeping the `node_modules`-above property they were put there for.

The near-miss row that made this safe was already in the table before it was the reason: `/p/.uf-notes/x.js` is Flow, because the rule matches the `.uf` path *segment*. `.uf-test-projects` is the same shape.

Locally green now: `assets` 12, `vite` 53, `cli` 69, `coverage` 11, `env_files` 19, `i18n` 7, `permissions` 5, `uf_transform` 84, and the three-way contract test.

ubugeeei:
Closing this: it duplicates a fix that landed in #643, with a worse mechanism, and the failing check is #643's guard catching exactly that.

#643 added `packages/vite/internal/compile-assets.js`. The payload enters the graph under a **NUL-prefixed synthetic id** rather than by path, so `isFlowModule`'s *first* rule declines it — the one about ids a build tool synthesised. That is better than what I did here, because it is about what the module *is* rather than where it happens to sit.

`tests/library/compile-assets.test.js` is the guard, and its last assertion is the one I broke:

```js
// And the file it comes from would *not* have been declined, which is why
// the id has to differ from the path. This is the assertion that fails if
// somebody puts the path back into the entry's import.
expect(isFlowModule("/app/.uf/build/compile/assets.js")).toBe(true);
```

That is deliberate. Leaving the path Flow-eligible is a **tripwire**: if a future change imports the payload by path again, the ceiling error comes back loudly instead of the file quietly slipping through unparsed. My `.uf/` exclusion would have removed it, so the regression this issue is about could return silently. The failing test was not an obstacle to the change — it was the change being wrong.

The file header says it plainly, and it quotes the same error I filed #679 about:

> The fix is which name the payload enters the graph under… Raising the ceiling would have weakened it everywhere to make room for the one file it should never have been asked about.

I filed #679 and built this at 12:0x while #643 was in the queue with the fix in it; it merged at 12:27Z. I should have checked whether an open PR already covered it before filing — this is the second duplicate I have made today, after #666 against #652.

Nothing here is worth salvaging. The fixture relocation to `.uf-test-projects/` in the second commit was only needed *because* of the exclusion, so it goes too.

Closing #679 as fixed by #643.

Issue #620, opened 2026-09-07
server: `uf preview` answers a draft request from a prerendered document


ubugeeei:
`uf preview` serves a prerendered document to a request carrying the draft-mode
cookie, so draft mode appears to be off there while it works under `uf start`
and `uf dev`.

`createStaticHandler` skips `.html` for a draft request (#615), but under
`uf preview` Vite's own file middleware runs before anything uf mounts behind
it, so that skip never sees the request.

This is the same shape as #342: a middleware order that puts a static file in
front of a decision uf has already made. The two should probably be fixed
together, or at least by one argument.

## Expected

The three servers agree. A request with a valid draft cookie is never answered
from a prerendered document, under `uf dev`, `uf preview` or `uf start`.

## Acceptance criteria

- A test that drives all three and asserts the same answer, so a fourth server
  cannot be added without answering the question.
- The reason a static file may not answer a draft request is written down once,
  not three times.

Issue #550, opened 2026-09-07
server: static serving resolves inside the root and then follows a symlink out of it


ubugeeei:
`packages/server/node.js` does the containment check the right way round — decode, reject NUL, `path.resolve`, then require the result to be the root or under it:

```js
const resolved = path.resolve(rootDir, `.${pathname}`);
if (resolved !== rootDir && !resolved.startsWith(rootDir + path.sep)) return null;
```

That defeats `../`, encoded `../`, and a NUL truncation. It does not defeat a **symlink**: the path is textually inside the root, `statFile` follows the link, and the bytes come from wherever it points.

`dist/` is written by the build, so the interesting source is `public/`, which is author-supplied and copied verbatim — and a dependency's `postinstall` writes into the project too. A `public/x` pointing at `/etc/passwd`, or at `.env`, is served with a 200.

This is the class of [CVE-2026-82393](https://nvd.nist.gov/vuln/detail/CVE-2026-82393) and [pnpm GHSA-6x96-7vc8-cm3p](https://github.com/pnpm/pnpm/security/advisories/GHSA-6x96-7vc8-cm3p) that `docs/security.md` already answers for the package store — the store is content-addressed so extraction cannot depend on attacker text — arriving at the other end of the toolchain, where the answer has not been given yet.

**Done**: the file that is opened is verified to be inside the root *after* symlinks are resolved, not before — `fs.realpath` on the candidate, compared against the realpath of the root — on every host that serves files from disk (`node.js`, `standalone.js`, and the deploy adapters that serve their own static half). A symlink inside the root that points inside the root still works, because that is a thing people do on purpose; one that points out is a 404. With a test that plants one.

Found while auditing uf against the Next.js and React CVE history.

Issue #391, opened 2026-09-06
deploy: six adapters, now that there is a seam and one implementation of it


ubugeeei:
`uf build --adapter node` writes a directory that runs on a host with a
JavaScript runtime and nothing else, and the half of it that is not Node is a
package export now. This issue is the other six adapters, and it exists so the
next one starts from the design rather than from the question.

## What is already there

**`@uniflowed/server/fetch` — `createFetchHandler({ app, document })`.** The
application as `Request` → `Response`: middleware, then the dispatcher for any
method, then a render for whatever is left. It touches no filesystem, holds no
Node types, and imports nothing from `@uniflowed/vite`. This is the whole of
what an adapter reuses. `uf preview`, `uf start` and `uf build --adapter node`
all answer through it — `packages/vite/internal/serve.js` calls it rather than
holding a copy — so a new adapter inherits the ordering decisions instead of
re-making them.

**`@uniflowed/server/node` — `createStaticHandler`, `nodeListener`, `serve`.**
Everything host-specific: a file out of a directory, the Node request
translation, the socket. An adapter for another host replaces this module and
nothing else.

**`uf_config::DeployAdapter::is_implemented`** is the single place the
Implemented/Planned split is recorded. `uf build --adapter edge` is an error
that names the target and this issue; `uf explain build` names the adapter that
will run; `app.runtime.deploy.adapters` defaults to what uf can actually
produce, which is `["node"]`.

**What `uf build --adapter node` emits**, for reference:

```
.uf/deploy/node/
  handler.js      the application: `export default { fetch }`, everything linked in
  server.js       node:http around it, serving ./static
  chunks/         the route modules, which are lazy import()s
  static/         a copy of dist/
  package.json    {"type": "module"}
```

## What each of the six still needs

The application is done for all of them. What is left, per target, is an entry
file and an answer to *where the static half lives* — and the second one is the
part with a decision in it.

**`bun` and `deno`.** The smallest of the six: `handler.js` already runs on
both, so each needs a `server.js` written against `Bun.serve` or `Deno.serve`
plus their own static answer. Worth doing only if the native servers measurably
beat `node:http` under the same handler — the `node` output already runs
unchanged on both, so an adapter that is not faster is a directory with a
different name on it. Measure first.

**`edge`** (Cloudflare Workers, Deno Deploy, Vercel and Netlify edge
functions). `handler.js` is already the artefact these want. Two things are
genuinely unresolved:

[...]

PR #689, opened 2026-09-08
feat(server, vite, cli): the bun deploy adapter, and the one line that makes it one

ubugeeei:
`uf build --adapter bun` writes the directory `bun server.js` runs. Part of #391.

It is the `node` output with `@uniflowed/server/bun` in place of
`@uniflowed/server/node`, and nothing else differs — `handler.js` is
byte-for-byte the same file, which
`every_adapter_answers_exactly_what_the_node_adapter_answers` now asserts
across five adapters instead of four.

## Why it exists

`DeployAdapter::unimplemented_because` said this was worth writing "only once a
benchmark shows the native server beating `node:http` under the same handler",
and beside it: *nobody has run that benchmark.* It has now been run, on the
same emitted `handler.js`, and Bun wins both halves of a request:

| path | node | bun | |
| --- | ---: | ---: | --- |
| dynamic — `/api/health`, c=32 | 14,006/s | 18,480/s | **+32%** |
| static — 215 KB asset, c=32 | 5,742/s | 7,251/s | **+26%** |
| static — 508 B asset, c=64 | 18,125/s | 27,862/s | **+54%** |

## Reuse the policy, swap the body

The static half decided the design, and it is the interesting result here.
`Bun.serve` handed a `Readable.toWeb(createReadStream(…))` — which is what
reusing `node.js`'s handler verbatim would give it — measures **5,670/s** on
that 215 KB asset at c=64 against `node:http`'s **5,732**. Level, inside the
noise. The entire win is `Bun.file`.

So `createStaticHandler` split in two:

- **`packages/server/internal/static.js`** — the policy. Containment, the
  symlink re-check (#550), the draft-cookie rule (#620), the
  `/guide` ⁄ `/guide/` ⁄ `guide.html` candidate order, the closed content-type
  table, and the NUL rejection. Shared.
- **each host** — one line for the body. `Readable.toWeb(createReadStream(…))`
  on Node, `Bun.file(…)` on Bun.

Re-implementing that policy against a second file API is how a third traversal
bug gets in; two issues have already been spent getting it right, and the
fastest `Bun.file` in the world does not pay for that.

`bun`'s bundle deliberately keeps `node`'s resolution conditions rather than
asking for Bun's own export condition: the win is measured in those two calls,
not in resolution, and a `handler.js` that stays identical is what makes an
adapter one file rather than one application. `edge` pays that price because it
must — there is no `node:stream` in a Worker.

## Tests

- `every_adapter_answers_exactly_what_the_node_adapter_answers` and
  `every_adapter_answers_the_same_server_action_call` both grew a `bun` row —
  five adapters, byte-identical answers.
- `the_bun_adapter_writes_a_directory_bun_serves_from_an_empty_one` runs the
  emitted `server.js` **on Bun**, from a directory with no `node_modules`
  above it, and asks it through the same function `uf preview` and `uf start`
  are asked with. It fails rather than skips when Bun is absent, like the two
  `--compile` tests; CI installs Bun for the `Test` job already.
- `assert_artefact_shape` refuses a `bun` bundle that carries `node:http`'s
  server or a `node` bundle that carries `Bun.` — the "directory with a
  different name on it" #391 warns about, asserted rather than assumed.
- `every_written_adapter_names_the_entry_it_writes` — a new contract test
  after `uf explain build` was caught saying a bun build writes "nothing yet".

Verified: `cargo test -p uf_cli -p uf_config` exit 0 (including the 54-test
vite suite), `cargo clippy --all-targets -- -D warnings` and `cargo fmt
--check` clean, 28 library tests green, and the emitted directory driven by
hand under Bun — static asset, prerendered document, route handler, rendered
page, `HEAD` with no body, and a `..%2f` traversal answered 404.

`deno` is still unwritten, for the reason `bun` was: nobody has run that
benchmark, and this machine has no Deno to run it with.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/689"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791489125&installation_model_id=422833&pr_number=689&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F689&signature=4b60d7f17de873cd27a8cd5aa30ba1f769c8d9d17a5206705e95402f16a2f87d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added Bun as a supported deployment target.
  - Bun deployments now generate and serve applications using `handler.js` and `server.js`.
  - Added static file serving for Bun, including content types, HTML fallback, and `HEAD` request support.

- **Bug Fixes**
  - Improved shared static-file handling for Node and Bun, including safer path validation and draft-page behavior.

- **Documentation**
  - Updated deployment documentation with Bun support, streaming, process persistence, and benchmark information.
  - Clarified that Deno remains unsupported.


ubugeeei:
Pushed 0fa51457 — the `Library` job was failing on this:

```
✗ published-packages.test.js  every package publishes every file it exports
    expected ["@uniflowed/server exports ./bun.js"] to equal []
```

`packages/server/package.json` gained `"./bun": "./bun.js"` under `exports`
and not `bun.js` under `files`, so the tarball would have shipped an export
pointing at a file that is not in it. `internal/static.js` needed nothing —
`internal/*.js` covers it.

Auto-merge (squash) is on.

Issue #348, opened 2026-09-06
rsc: client-only APIs are matched by name, so a hook that wraps one is invisible to the graph


ubugeeei:
`rsc/client-only-api-in-server` matches identifiers against two sorted name
lists in `crates/uf_rsc/src/scan.rs` — `CLIENT_ONLY_APIS` (React's hooks) and
`CLIENT_ONLY_GLOBALS` (`window`, `document`, `localStorage`, …). A call to
`useState` in a Server Component is caught. A call to *anything that wraps it*
is not.

This repository has the case in it. `docs/app/_uf.layout.js` is a Server
Component by classification, and its `Masthead` calls `useRoute()` from
`@uniflowed/router` — a hook built on `useContext`, which is on the list. The
graph reports nothing, because the name `useRoute` is not. The same is true of
every hook `@uniflowed/hooks`, `@uniflowed/state` and `@uniflowed/form` export,
and of every hook a user writes.

It did not matter while nothing was split — every module runs in both places
today — and it is the moment the split lands (#252) that a false negative here
becomes a page that renders on the server and throws.

## Why the name list is still the right starting point

It has no false positives worth the name: the scanner skips declaration sites,
skips property accesses (`foo.document`), and requires a call for the hook
names, so a module that defines its own `useState` is not reaching for React's.
That is worth keeping.

## What would close it

The information exists in this repository already, in two places:

* **The graph.** A module whose exported hook transitively calls a client-only
  API is itself client-only. `RscGraph` has the import edges and
  `scan_exports` has the exported names; what is missing is the binding between
  an export and the APIs its body reaches, which is a per-module fact the scan
  could record.
* **The React Compiler.** `react/no-render-side-effects` already decides
  whether a component's renders are repeatable, over a real AST rather than a
  token stream — which #305 argues is the same property that decides whether a
  component can run on the server. `crates/uf_react_compiler` computes it and
  the RSC classification does not read it.

Either is a larger change than the rule it fixes, which is why this is written
down rather than attempted alongside #281. The honest statement for now is the
one `docs/architecture.md` makes: the contract is enforced against a split that
has not happened, and its coverage is name-based.

PR #690, opened 2026-09-08
feat(rsc, lib): the hooks the registry already knew about are decided, not asked about

ubugeeei:
## Summary

`rsc/unclassified-hook-in-server` says the analysis stopped. For 57 hooks it did
not have to: `uf_lib`'s `hook_descriptors()` has carried `server_component_safe`
for every hook `@uniflowed/hooks` exports since before `uf_rsc` existed, and
**nothing read it**. A Server Component calling `useMediaQuery` got a warning
saying uf could not say, from a binary that could.

`rsc/client-only-hook-in-server` is the rule that reads it — an error, with the
package named:

```
server module `app/page.js` calls `useMediaQuery` at line 2:44, and
`@uniflowed/hooks` exports it as a hook that only runs in the browser
```

Same verdict `rsc/client-only-api-in-server` gives `useState`, for the same
reason: `useMediaQuery` reads `matchMedia` through a `useSyncExternalStore`, and
a server has neither. `useIsomorphicLayoutEffect` — the one row in that table
that *is* safe — stays out of the rule; reading a table without believing it is
worse than not reading it.

## Why one package, when eleven export hooks

#388 puts the data change first: a `client_only` set per registered module. Done
that way it is 94 hand judgements (`hooks` 58, `form` 7, `state` 7, `router` 5,
`query` 4, `relay` 4, `tui` 3, `web` 3, one each from `loader`, `rm`,
`validator`), and getting one wrong turns a warning into an **error** in
somebody's working build. Deriving them instead needs to attribute a body to an
export — which is the fact #388's *other* half adds. I tried the derivation and
it calls `useGeolocation` and `useStorage` server-safe, which they plainly are
not, for exactly that reason.

`@uniflowed/hooks` is the one package that needs neither: its table exists, and
`the_hook_table_names_exactly_the_hooks_the_package_exports` (#683) holds it to
the package's own sources. So this is the *plumbing* — how a hook call is
attributed to the package that exports it — proved against data somebody has
already checked. The other ten follow when the fixpoint can generate them.

## What it deliberately is not

Attribution is through the module's imports, not the binding the import
introduced: `ImportSpecifier` keeps the specifier and drops the names. So a
module that imports `@uniflowed/hooks` **and** shadows one of its names is
decided wrongly; a module that never imports it is not decided at all. Both are
asserted, the second one because a project writing its own `useMediaQuery` lands
there.

`uf_rsc` gains a dependency on `uf_lib` rather than copying 57 names into a
second list — that copy is exactly the drift #683 fixed, one directory over.

## Blast radius

None on uf's own site. `analyze_project` reads `.js` outside `.uf`, `dist` and
`node_modules`, and the four `@uniflowed/hooks` mentions under `docs/` are
inside MDX prose. The four call sites #348 found are `useRoute` and `useTheme`,
which stay unclassified.

## Verification

- [x] `cargo test -p uf_rsc` — 212 passed (was 207; six new, one of the new
      tests replaces nothing)
- [x] `cargo test -p uf_lib` — 32 passed
- [x] `cargo clippy -p uf_rsc --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`

Part of #388.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791490330&installation_model_id=422833&pr_number=690&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F690&signature=f2b789c2caa07e254a83a70af3597e150cb38defa6fc1118851de45ef11166c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Verified against this repository's own site, which is the case #348 was filed
about. `tools/docs/build.sh` on this branch:

```
app/_design/parts.js      1 warning   rsc/unclassified-hook-in-server  useRoute
app/_uf.layout.js         2 warnings  rsc/unclassified-hook-in-server  useRoute, useTheme
app/guide/_uf.layout.js   1 warning   rsc/unclassified-hook-in-server  useRoute
✓ build succeeded in 6.55s
```

Four warnings, zero `rsc/client-only-hook-in-server`, build green — exactly the
four sites #388 names, and they stay questions because `useRoute` is
`@uniflowed/router`'s and `useTheme` is the site's own. Blast radius as
claimed: nil.

That run also turned up a defect in the neighbouring rule, fixed in 110f708a.
The message and the caret disagreed about the same call:

```
warning[rsc/unclassified-hook-in-server]: … calls `useRoute` at line 122:24, …
  --> app/_uf.layout.js:122:1
```

`UnclassifiedHookInServerModule` has carried a column since #348 and `column()`
had no arm for it, so every reader got the fallback that means "the whole line
is at fault" — and `uf build` passes `diagnostic.column()` straight to the
snippet renderer. Asserted now, on the number rather than on "greater than 1".

Issue #531, opened 2026-09-07
server: scheduled work, and where a cron runs when the deployment has none


ubugeeei:
A cron is the other half of the queue in #NEXT: work that no request starts.

The difficulty is that the deployment targets disagree completely. A `node` or `container` adapter can hold a scheduler in-process. A `serverless` one cannot and needs the platform's scheduler to call a route. An `edge` one is the platform's scheduler.

**Done**: a schedule declared in the application — a route module that exports a cron expression and a handler — and an adapter contract that either runs it or emits the platform's own configuration for it. A target that can do neither must refuse the configuration at build time rather than build something that never runs, which is the failure this feature has everywhere it exists.

Part of #488, and adjacent to #391.

ubugeeei:
**Half of this landed** in [#696](https://github.com/ubugeeei-prod/uf/pull/696): `@uniflowed/server/schedule`, and the refusal.

## What exists now

`defineSchedule({ name, cron, run })` and `createScheduler`. uf owns *when* — five fields, UTC, POSIX's either-day rule, and running a minute at most once however often the scheduler is ticked. `node` and `bun` take `schedules` on `serve` and tick them through one shared helper; ticking stops when `close` resolves.

And the refusal this issue asks for by name. `ServerCapabilities` gained `scheduler`, and `assertCapable` refuses one whose `triggered` is false on a target that keeps no process:

> nothing would be running at the minute the schedule names: this target keeps no process, and its platform has not been told to call one

So `serverless` and `edge` fail where the host is wired rather than building a deployment whose scheduled work silently never runs — which is the failure this issue says the feature has everywhere it exists.

The syntax is deliberately only what every cron agrees on: `*`, `n`, `a-b`, `*/n`, `a-b/n` and comma lists. No `@daily`, no `JAN`/`MON`, no `L`/`W`/`#`. Review caught `5/15` being accepted with the step dropped — an hourly-looking schedule that ran once a day — and it is refused now with the working spelling in the message.

## What is left, and the shape it wants

The declaration and the emission — the two halves that need the build rather than the runtime:

[...]

PR #712, opened 2026-09-09
feat(cli): a schedule declared in a route handler, and the build that refuses to ignore it

ubugeeei:
The build's half of [#531](https://github.com/ubugeeei-prod/uf/issues/531). [#696](https://github.com/ubugeeei-prod/uf/pull/696) shipped the runtime — `defineSchedule` and a scheduler a process can tick — and left the declaration and the emission, which need the build rather than a running server.

A route handler says when it should run without being asked:

```js
// app/api/sweep/_uf.route.js
export const schedule = "*/15 * * * *";

export function GET(): Response {
  return Response.json({ swept: true });
}
```

## Why a string and not a call

`uf build` writes a platform's cron configuration **without running a line of the project**, so an expression it would have to evaluate is one it cannot write down. `export const schedule = everyMinutes(15)` is refused by name rather than skipped, because a schedule the build could not read is a schedule the deployment would not run — which is what this issue exists to refuse.

## What each adapter does

`--adapter edge` puts it where Cloudflare's own scheduler reads it. Verified against the real build:

```json
"triggers": { "crons": ["*/15 * * * *"] }
```

Everything else **refuses the build**, and names what it found:

```
error: the `node` adapter would not run the 1 schedule(s) this project declares,
so this build would produce a deployment whose scheduled work never happens:
    /api/sweep — `*/15 * * * *`, in app/api/sweep/_uf.route.js
  `--adapter edge` writes them into `wrangler.json` for Cloudflare's own
  scheduler. On a target that keeps a process, pass them to `serve` yourself
  with `@uniflowed/server/schedule` — declaring one in a route module is not
  wired for those yet (ubugeeei-prod/uf#531).
```

That is the clause #531 asks for by name: *"a target that can do neither must refuse the configuration at build time rather than build something that never runs, which is the failure this feature has everywhere it exists."*

`node`, `bun` and `container` keep a process and *could* run one — the entry uf generates for them does not pass the declaration to `serve` yet, so today they would not, and saying so beats the silence. That remains #531's.

The refusal runs immediately after the adapter is resolved, before the bundler does anything, so a project whose scheduled work would never happen hears about it in a second rather than after a build.

## What this deliberately does not check

What a field *contains*. `packages/server/internal/cron.js` owns that, and a second matcher here would be two implementations of one rule, drifting the first time either learned something. So `99 * * * *` passes the build with five fields and is refused by `defineSchedule` when the deployment starts — loudly, rather than quietly accepted in both places. There is a test asserting exactly that, so the split is deliberate rather than an oversight.

Reading is through `uf_transform::lowered_ast`, the same parser the build transforms with, rather than a third opinion about what a module is.

## Verification

- 13 unit tests for discovery — string forms, a real project tree, middleware excluded, path ordering, and every refusal.
- 4 for the emission and the refusal across all seven adapters.
- End to end against the real binary: `--adapter node` refuses in a second; `--adapter edge` builds and writes the `triggers.crons` above.
- `cargo test -p uf_cli` exit 0, 462 lib tests; `clippy --all-targets -- -D warnings` and `fmt --check` clean; docs tests green.

Rebased onto current main after #697, #702, #703 and #704 landed mid-work.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/712"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791525954&installation_model_id=422833&pr_number=712&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F712&signature=d3408d1b1d8586c900322bd138c5995df7ad8c118091cbee60469fb312eee780"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Route handlers can declare cron schedules using string literals.
  - Edge deployments now include declared schedules in the generated platform configuration.
  - Schedule declarations are discovered automatically and validated during deployment.

- **Bug Fixes**
  - Builds now fail clearly when schedules are used with unsupported deployment adapters, identifying affected routes and files.

- **Documentation**
  - Added routing guidance for declaring schedules and documented schedule support for the Edge adapter.

### DIFF
--- a/crates/uf_cli/src/commands/build.rs
+++ b/crates/uf_cli/src/commands/build.rs
@@ -275,6 +275,17 @@ pub(crate) fn build(
     // which has a bundle to load and does not have it; that is refused in
     // `commands::serve`, where it is a fact rather than an opinion.
     let adapter = deploy::resolve(&resolved.config.app.runtime.deploy, requested_adapter)?;
+    // Before the build rather than after it: a project whose scheduled work
+    // this target would never run should hear so in a second, not after a
+    // bundle. ubugeeei-prod/uf#531.
+    let declared_schedules = match adapter {
+        Some(adapter) => {
+            let found = deploy::schedules::discover_schedules(&root, &resolved.config)?;
+            deploy::schedules::refuse_unrunnable(adapter, &found)?;
+            found
+        }
+        None => Vec::new(),
+    };
     // The fourth thing that needs a process, and the only one `uf` can see
     // without evaluating a module. Checked here rather than in the builder for
     // exactly that reason — see `refuse_unanswerable_actions`.
@@ -486,7 +497,9 @@ pub(crate) fn build(
                 "writing the {} adapter's output",
                 adapter.as_str()
             ));
-            Some(timer.measure("adapter", || deploy::deploy(ui, adapter, link))?)
+            Some(timer.measure("adapter", || {
+                deploy::deploy(ui, adapter, link, &declared_schedules)
+            })?)
         }
         None => None,
     };
--- a/crates/uf_cli/src/commands/deploy.rs
+++ b/crates/uf_cli/src/commands/deploy.rs
@@ -88,6 +88,7 @@ use crate::commands::vite::{Driver, Event, LinkContext, LogLevel, render_error, 
 use crate::support::project_label;
 use crate::ui::Ui;
 
+pub(crate) mod schedules;
 pub(crate) mod static_host;
 
 /// Where the generated entry files are written before they are linked.
@@ -201,6 +202,7 @@ pub(crate) fn deploy(
     ui: &mut Ui,
     adapter: DeployAdapter,
     link: LinkContext<'_>,
+    schedules: &[schedules::DeclaredSchedule],
 ) -> Result<Deployed> {
     let LinkContext {
         host,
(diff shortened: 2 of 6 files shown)


ubugeeei:
Fixed and pushed. `export { schedule }` carries `"declaration": null` — present, and not a declaration — so `if let Some(declaration)` took that branch and `continue`d straight past the specifiers. Both spellings now reach the same answer, `export { every as schedule }` resolves through the local name, and a re-export from another module is refused rather than missed.

Finding the gap also turned up that the `mod tests;` line had been lost in an earlier edit of mine: thirteen tests had stopped running and the suite still said "ok". The count was what caught it — 449 where 462 was expected.

Issue #316, opened 2026-09-06
tui: uf's own CLI still draws with a private renderer, and cannot reach the public one


ubugeeei:
ubugeeei-prod/uf#247 asked for two things. `@uniflowed/tui` is a real renderer now; the other half — **uf's own CLI as its first user** — is not done, and the reason is structural rather than a shortfall of effort, so it is written down here rather than left to be rediscovered.

## Why it did not land

`crates/uf_term` is Rust and is what every `uf` command draws with. `@uniflowed/tui` is Flow-typed JavaScript. For the CLI to draw through it, one of these has to be true and none is:

1. **A JavaScript-to-native bridge**, so Rust can drive the JS renderer. There is none in this repository — every `nativeRuntimeRequired` is a promise of a boundary that does not exist — and driving a per-frame renderer across a Node process boundary is the wrong shape anyway: a banner would cost a Node start-up.
2. **The renderer in Rust**, with the JS package as a binding over it. That was the decision ubugeeei-prod/uf#247 turned down; `packages/tui/index.js` says why, and the argument does not change because of this issue.
3. **A uf command written in Flow.** There is no way to *run* a Flow program in this repository outside `uf test`, `uf dev` and `uf build`. `uf run` runs shell commands, and Node cannot load a Flow module without uf's transform.

So the honest position today is that the two renderers are separate and each is right for its own caller: `uf_term` for a Rust CLI whose start-up budget is a few milliseconds, `@uniflowed/tui` for an application somebody writes.

## What would actually close it

The third option is the interesting one, and it is a product feature rather than a workaround: **a way to run a Flow entry point on the Capability JS Host** — the mechanism `uf lint` plugins already use (`crates/uf_plugin`) pointed at a `render()` call instead of a rule. With that, a uf command that is genuinely a terminal *application* rather than a line of output — `uf test --watch`, which is the strongest candidate — could be a Flow module, and the dogfooding argument in `ubugeeei-redundancy.md` would hold for the TUI the way it already does for `uf fmt` and `uf lint`.

That is a large piece and it should be measured before it is committed to: a watch-mode UI that costs a Node start-up on every run is worse than the Rust one it replaces.

## In the meantime

The two renderers must not silently disagree about the terminal. Two things already stop that and both should stay:

- `packages/tui/capability.js` reproduces `crates/uf_term/src/capability.rs`'s precedence order variable for variable, and its header says the list is that one.
[...]

Issue #315, opened 2026-09-06
tui: measure `@uniflowed/tui` against React Ink, or stop mentioning it


ubugeeei:
`crates/uf_tui/src/lib.rs` used to carry `performance_target: "faster-than-react-ink"`, which was never measured. ubugeeei-prod/uf#247 replaced it with `"writes-only-changed-cells"`, which is measured — `tests/library/tui.test.js` asserts that in an 80×24 terminal the first frame sends 1,920 cells and changing one character of a status line then sends **one cell in seven bytes** — but that is a measurement of one renderer, not a comparison.

The comparison is still worth having, because Ink is what a reader is coming from and because the mechanism is a real difference rather than a language one: Ink renders to a string and reprints from the first changed line to the bottom of the frame, so the same edit costs it however much of the screen is below it.

## What this needs

- Ink installed as a dev dependency of the benchmark, not of `@uniflowed/tui`.
- A workload both libraries can render, written twice, with the same output: a list that scrolls, a status line that ticks, and a full-screen redraw. Ink has no `Input` equivalent to `@uniflowed/tui`'s, so the input path is out of scope.
- The numbers `ubugeeei-redundancy.md` requires of a performance claim: hardware, versions, workload, cache state, and variance — not one run of each.
- Both metrics, because they are different claims: **bytes written per frame** (the mechanism) and **wall-clock time from a state update to the write returning** (what a person feels).

## Where the numbers go

`docs/app/guide/tui/_uf.page.mdx` already carries the uf-only measurement in a table, the way `docs/app/guide/format/_uf.page.mdx` carries the formatter's, and `tests/library/tui.test.js` asserts the table's figures against the renderer so the page cannot go stale. Extend both.

If the comparison turns out not to favour uf, that is the answer and it goes on the page. An honest slower is worth more than an unmeasured faster, which is what this replaced.

Issue #314, opened 2026-09-06
tui: the half of OpenTUI `@uniflowed/tui` does not implement yet


ubugeeei:
`@uniflowed/tui` is a real renderer now (ubugeeei-prod/uf#247): a component tree, flexbox in whole cells, a cell buffer, a diff that writes only what changed, keyboard input with OpenTUI's names and propagation rules, declarative focus, capability detection, and an in-memory renderer. Three components: `Box`, `Text`, `Input`.

OpenTUI has considerably more, and the contract in `crates/uf_tui/src/lib.rs` now says so by *omission* rather than by listing features nothing implements. This issue is the list.

## Input

- **Mouse.** `TuiInputModel` is `keyboard-focus` because mouse reporting is not enabled and no event is routed. It needs the `?1000h`/`?1002h`/`?1006h` mode set, an SGR-1006 decoder, a hit grid written during `paint`, and the bubble-and-capture rules in [OpenTUI's interaction page](https://opentui.com/docs/core-concepts/interaction) — `over`/`out` on hit-target change, left-drag capture, `drop` carrying `event.source`.
- **Selection.** Drag selection over `selectable` text, one selection per renderer, and `getSelectedText()`.
- **Bracketed paste.** `keys.js` decodes single keys; `ESC[200~ … ESC[201~` is not recognised, so a paste currently arrives as a burst of individual keys.
- **Key release.** `KeyEvent.eventType` is always `"press"`. Release needs the Kitty keyboard protocol.

## Layout

`packages/tui/layout.js` says at the bottom what it does not do, and all three are here:

- `flexWrap` (OpenTUI defaults to `"no-wrap"`, so this is not urgent).
- `position: "absolute"`, which needs a containing-block concept the package does not have.
- `auto` margins.

## Components

`Select`, `TabSelect`, `Slider`, `Textarea`, `ScrollBox`, `ScrollBar`, `Code`, `Markdown`, `LineNumbers`, `Diff`, `TextTable`, `AsciiFont`, `FrameBuffer`, `Image`, `QrCode`, `EmbeddedTerminal` — and the application APIs: clipboard, notifications, audio, `Timeline`.

`ScrollBox` is the one worth doing first, and not because it is easy: it is where the "only the visible window is laid out and diffed" property has to be proved, and that property is what keeps a ten-thousand-line log cheap.

## What done looks like

[...]

PR #721, opened 2026-09-09
feat(tui): a scrolling box that costs its window, not its content

ubugeeei:
The first slice of #314, which names `ScrollBox` as the one to do first and says
why: it is where *only the visible window is laid out and diffed* has to be
proved. A `ScrollBox` that lays out all its children and clips is not that
feature.

Half of it was already true. #449 stopped laying out and painting children
outside the window, and `TuiFeature::Scrollback` was claimed on that basis. The
other half was not: every frame still **measured** every child, called `hide()`
on every child it skipped, and walked past every child in the painter. A
hundred thousand rows cost a hundred thousand of each, three times per
keystroke.

## The measurement

Two of them, deliberately different in kind — a count cannot tell you a constant
is small, and a clock cannot tell you a curve is flat.

**Counted, in `tests/library/tui.test.js`.** The suite builds `LayoutNode`s by
hand, which is what `layout.js` is designed for, so it can count the calls
layout makes into its leaves. A measurement is a call *into* a leaf and nothing
above layout can watch one happen.

| | measure calls | children laid out |
| --- | --- | --- |
| 100,000 rows, first frame | 100,000 | 24 |
| ...then scrolled to row 1 / 500 / 50,000 / 99,976 | **0** each | 24 each |
| ...then one line appended | **1** | 24 |
| 30 rows, same window | — | 24, the same indices |

Four of the five new cases **fail against the renderer as it is on `main`**,
with `expected 100000 to be 0` and `expected 100001 to be 1`. They are not
descriptions of the change; they are the discriminator.

**Timed, in `tools/bench/tui/window.js`** (`uf run bench:tui:window`, new).
Layout + paint + diff only, 80×24, median of 20 keystrokes, range over three
whole runs:

| rows | first frame | one row scrolled | one line appended |
| --- | --- | --- | --- |
| 30 | 0.33 ms | 0.16 – 0.29 ms | 0.17 – 0.32 ms |
| 1,000 | 5.8 ms | 0.20 – 0.28 ms | 0.16 – 0.41 ms |
| 10,000 | 65 ms | 0.21 – 0.25 ms | 0.23 – 0.28 ms |
| 100,000 | 602 ms | 0.21 – 0.31 ms | 0.26 – 0.33 ms |

The thirty-row band and the hundred-thousand-row band overlap: three thousand
times the content, and the frame cannot tell. **Before**, the same scroll cost
0.23 ms, 3.64 ms, 59.64 ms, 491.25 ms — a straight line through the origin. At
a hundred thousand rows that is 491 ms → 0.24 ms.

The first-frame column is still that line, deliberately: the height of the
content is what `Number.MAX_SAFE_INTEGER` gets clamped against, so every row is
asked its height once. It is also the noisy column, and is reported as the
fastest of three for the reason `startup.js` gives.

The benchmark prints a second table that is **not** this component's, because
publishing only the flat one would be publishing half of it: React reconciling
the tree costs ~0.2 ms at thirty rows and a few hundred at a hundred thousand.
A hundred thousand `<Text>` children are a hundred thousand fibers whoever is
scrolling them, and a `ScrollBox` cannot render children the caller did not
create.

## Where the layout/diff boundary went, and why

**In JavaScript, in `packages/tui`.** Not a preference — `crates/uf_tui/src/lib.rs`
opens with *"This crate holds no renderer"*, and the argument against a Rust
core with a binding is written out in `packages/tui/index.js`. The Rust crate is
the contract `uf inspect` prints; putting the windowing behind it would need a
JS-to-native bridge uf does not have, and would have to cross that bridge on
every frame to answer a question about a tree React owns. Nothing in the Rust
half changes here except one doc comment.

The same reasoning settles the measurement's home. `crates/uf_profiler` counts
allocations in Rust and criterion benches Rust crates; neither can see a
JavaScript renderer. The repository already measures this package in
`tools/bench/tui/` (#315's Ink comparison, #316's start-up), so the new bench
went beside them, and the *exact* half went into the test suite where it runs in
CI on every machine — a count is the same number everywhere and a millisecond is
not.

Inside JS, the boundary that mattered was between **React and layout**. Layout
cannot know that a tree changed without comparing it with the one it saw last,
which is the walk the whole change exists to avoid; React knows exactly. So
`internal/tree.js` writes `scrollDirtyFrom` and layout reads it. It is a plain
number on the node rather than a call into `layout.js` because that file's own
header says the three participants do not import each other, and a node's fields
are the whole of what they say to one another.

## What changed

- `ScrollIndex` — a scrolling box keeps where each child sits, and rebuilds only
  the part that moved. An **append** dirties from the old child count, so a log
  that grew by a line re-measures a line.
- The first visible child is found by **binary search** over the stack.
- The painter is handed the **range** of children layout placed (`scrollFirst`,
  `scrollCount`) instead of a `hidden` flag per child, so the paint walk is the
  window too. `LayoutNode.hidden` is gone; it was only ever the mechanism for
  this.
- `intrinsicSize` keeps what it measured, keyed on what it was offered;
  `invalidate` clears it up the tree on every mutation React makes.
- A scrolling box's **own** props deliberately do not invalidate its stack: they
  cannot move its children except through the width, the viewport and the gap,
  and all three are in the stack's key. `scrollTop` is a prop on that box, so
  without that exemption every notch would invalidate the thing that makes
  scrolling cheap. The benchmark caught exactly that on its first run — 14 ms
  per scroll at a hundred thousand rows.
- Two renderer-level tests guard the kept measurement from the other side: a row
  **above** the window growing taller, and one going away, both move what is
  under them.

## The not-claimed list

**No line was deleted, and that is the finding.** The list in
`crates/uf_tui/src/tests.rs` is `Selection`, `Keymap`, `TerminalAutomation`,
`CodeHighlight`, `Markdown`, `Images`, `Audio`, `ThreeD`, `Ssh`, `QrCode`,
`EmbeddedTerminal`, `Clipboard`, `Notifications`, `Animations` — and
`Scrollback` left it in #449, before the property it names was true. This change
announces nothing new; it makes an existing claim honest. So the contract moved
the other way: `TuiFeature::Scrollback`'s doc comment now says which part is
proved, how it is counted, and which part — the first frame — is still
proportional to the content and why it has to be.

Selection is the next line to delete, and it is untouched here.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p uf_tui -p uf_term` — 5 + 252 pass, plus 2 doc-tests
- `uf run test:lib` — 2,772 pass, 0 fail (100 in `tui.test.js`, 21 of them
  `ScrollBox`)
- `uf fmt --check`, `uf lint` — clean, 0 errors

Docs: `docs/app/guide/tui/_uf.page.mdx` replaces "Every child is measured" with
what it costs now, both tables, the before numbers, and the part that is React's
rather than this renderer's. `docs/issue-roadmap.md` records the property beside
the scrolling entry.

Closes nothing — #314 is XL and this is one slice of it.

## Summary by CodeRabbit

- **Performance**
  - Improved `ScrollBox` performance for large content by limiting layout, painting, and updates to rows visible in the scroll window.
  - Added caching and incremental updates to reduce repeated measurement during scrolling and content changes.
  - Added support for efficiently handling row growth, removal, and appended content.

- **Documentation**
  - Expanded `ScrollBox` and Terminal UI performance guidance, including benchmark results and remaining application-side costs.

- **Tests**
  - Added coverage for visible-window layout, caching, scrolling, row changes, and measurement behavior.

### DIFF
--- a/crates/uf_tui/src/lib.rs
+++ b/crates/uf_tui/src/lib.rs
@@ -208,6 +208,17 @@ pub enum TuiFeature {
     /// other mouse event — but the offset is still a prop, so scrolling
     /// remains something the application does rather than something the
     /// component does to itself.
+    ///
+    /// What this word does promise, and what ubugeeei-prod/uf#314 asked to be
+    /// proved rather than asserted, is that the *window* is what a scrolling
+    /// box costs. Moving one over a hundred thousand rows measures nothing,
+    /// lays out the rows on the screen, and walks into no others; a renderer
+    /// that laid all hundred thousand out and clipped them would draw the
+    /// same frames and is not this. It is counted rather than claimed —
+    /// `tests/library/tui.test.js` counts the calls layout makes into its
+    /// leaves — and the one pass still proportional to the content is the
+    /// first frame, which has to ask every row its height once because the
+    /// height of the content is what the offset gets clamped against.
     Scrollback,
     /// Key binding and command routing.
     Keymap,
--- a/docs/app/guide/tui/_uf.page.mdx
+++ b/docs/app/guide/tui/_uf.page.mdx
@@ -222,12 +222,59 @@ does not, a list moves a selection and lets the box follow *that*. A component
 that owned the offset would also have to own "how far is a page", which is the
 viewport's height, which nothing knows until layout has run.
 
-**What it costs.** Every child is measured, because the total height is what the
-offset is clamped against. Only the children that intersect the window are laid
-out, walked into, or painted — so ten thousand rows in a twenty-four row terminal
-is one measure per row and a screenful of everything else. `tests/library/tui.test.js`
-renders exactly that and asserts the first frame costs the terminal's 1,920 cells
-rather than the log's, and that changing one visible row afterwards costs one.
+### What a window costs
+
+The window, and not the content. Moving the offset over a hundred thousand rows
+measures none of them, lays out and paints the twenty-four that are on the
+screen, and never visits the rest; appending a line to that log measures the
+line. The one pass that does grow with the content is the **first** frame, and
+it has to: the height of the content is what `Number.MAX_SAFE_INTEGER` is
+clamped against, so every row is asked its height once. After that the answer is
+kept until something under that row changes.
+
+That is counted rather than claimed. `tests/library/tui.test.js` builds the
+layout tree by hand so that it can count the calls layout makes into its leaves,
+and asserts a hundred thousand measurements on the first frame and **zero** on
+every scroll after it — a renderer that laid all hundred thousand rows out and
+clipped them would draw the same frames and fail that. Beside it, the same suite
+renders ten thousand rows through the real renderer and asserts that the first
+frame costs the terminal's 1,920 cells rather than the log's, and that changing
+one visible row afterwards costs one.
+
+The clock agrees, which is `tools/bench/tui/window.js` — `uf run
+bench:tui:window`. Layout, paint and diff, at 80×24; each cell is the median of
+twenty keystrokes, and the range is across three whole runs on the machine in
+the provenance note above, while it was shared with other work:
+
+| rows | first frame | one row scrolled | one line appended |
+| --- | --- | --- | --- |
+| 30 | 0.33 ms | 0.16 – 0.29 ms | 0.17 – 0.32 ms |
+| 1,000 | 5.8 ms | 0.20 – 0.28 ms | 0.16 – 0.41 ms |
+| 10,000 | 65 ms | 0.21 – 0.25 ms | 0.23 – 0.28 ms |
+| 100,000 | 602 ms | 0.21 – 0.31 ms | 0.26 – 0.33 ms |
+
+Read the last two columns as one number: at every size a frame lands in the same
+fifth of a millisecond, and the thirty-row band and the hundred-thousand-row band
+overlap. Three thousand times the content, and the frame cannot tell.
+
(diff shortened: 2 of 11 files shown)


ubugeeei:
Pushed `9c03e4ae` for the review finding on `packages/tui/internal/paint.js`. The thread auto-resolved on the push, so recording the answer here rather than leaving it implied.

**The finding was right.** `hideInstance` writes `width: 0, height: 0, hidden: true` — React's hide for a Suspense fallback and for `<Activity>` — and `paint` read none of it. The four existing mentions of `hidden` in that file are all `overflow: "hidden"`, a different property.

The zero size is not sufficient on its own: `overflow` defaults to `"visible"`, so a child laid out inside a box that has since collapsed to 0×0 still draws over its edge, and the walk would also record hits for a subtree the reader cannot see. It is the same hazard the scroll branch in this very PR guards against in its own words — *"their geometry is from whichever frame last showed them and drawing it would put those rows back on the screen"*.

The fix is one property read at the top of `paint`, before anything is drawn.

**It ships without a test, deliberately, and that is worth stating plainly.** The condition is stateful — a subtree laid out and *then* hidden, keeping the geometry of the frame that showed it. A static render of a `hidden` 0×0 box does not reproduce it, because layout gives the child no room either. I wrote that test, ran it with the guard disabled, watched it pass, and deleted it. A test that passes without the fix is cover rather than evidence, and this repository's habit is to say which of the two it has. Reproducing this properly wants a suspending transition over an already-laid-out subtree, and that is worth its own change.

`uf test#library`: 2772 passed, 0 failed, 1 skipped.

PR #728, opened 2026-09-09
feat(test): the browser as a host, so a component test runs where a component runs

### LINKED ISSUE #517: test: browser mode, so a component test runs where a component runs
`@uniflowed/react-testing` renders into happy-dom, which is fast and is not a browser: layout is approximated, `IntersectionObserver` is a stub, and anything that reads a computed size is untestable.

A browser mode runs the same test file in a real browser through the same runner, which is what Vitest's browser mode is for.

**Done**: `uf test --browser` runs the existing suite in a real browser; the runner still schedules in Rust and the browser is another host, which is the shape `uf_test` already has for Node and Bun; the same assertions pass in both modes or the difference is named. `docs/roadmap.md` already lists "Playwright-compatible browser automation" under P2 — this is the smaller half of it, and worth having first.

Part of #488.

ubugeeei:
Closes #517.

`@uniflowed/react-testing` renders into happy-dom, which computes no layout.
The expensive part is not that it refuses to answer — it is that it *answers*:

| Asked of a half-width box | happy-dom | a browser |
| --- | --- | --- |
| `getComputedStyle(el).width` | `"50%"` | `"300px"` |
| `el.getBoundingClientRect().width` | `0` | `300` |
| an `IntersectionObserver` on it | never fires | fires |

The first row is the one that costs something. happy-dom resolves the cascade
and stops, so the value it hands back is the *declared* one wearing the name of
the computed one. A test asserting that a panel is half its container's width
passes against a string that says nothing about width, and keeps passing after
the component stops being half its container's width.

`uf test --browser` runs the same files, with the same `describe` / `it` /
`expect`, in a real page.

## How the browser fits `uf_test`'s host seam

`HostKind::Browser`, beside `Node`, `Bun` and `Deno`. Nothing else about the
seam moves: the driver is spawned like any worker, is written to on stdin,
answers with the same events on stdout, is bounded by the same wall clock, and
is killed and replaced when it stops answering. `crates/uf_test/src/host.rs`
does not branch on which host it is talking to, and **scheduling stays in
Rust** — discovery, longest-first ordering from recorded timings, concurrency,
deadlines, retries and the report all work exactly as before.

The process uf starts is a Node driver (`@uniflowed/test/browser-worker.js`),
because a page cannot read a pipe. It serves the page its modules — through the
same `uf transform` a build uses, with the same options the Node loader passes —
and shuttles JSON between a pipe and a socket. Node whatever the project's
Capability JS Host is: the runtime under test is the browser, and pinning the
driver removes an axis from a feature that has enough of them. `docs/hosts.md`
carries that row.

**One page load per file**, which is stronger isolation than Node mode gets. A
Node worker serves many files out of one process, so everything two files could
share has to be put back by hand (`internal/isolation.js`, three issues long).
A page reloads: new realm, new document, new module registry, and a
`setInterval` the last file abandoned is gone rather than merely ignored.

## What it depends on, and how it refuses

**A browser that is already on the machine.** uf does not download one, does not
vendor one, does not pin a version, and links no driver library — there is no
Playwright and no Puppeteer here. `crates/uf_test/src/browser.rs` looks on
`PATH` for `chromium`, `chrome`, `google-chrome`, `google-chrome-stable`,
`microsoft-edge`, `brave-browser`, then for an installed Google Chrome, Chromium
or Microsoft Edge; `UF_BROWSER` names one outright and is never second-guessed.
That is `docs/red-lines.md` applied to a dependency heavier than a JS runtime:
uf owns orchestration, not implementation, and every built-in provider stays
replaceable by naming a binary.

The list is Chromium-shaped and says so. A run needs to tell a browser from a
command line to open a page headlessly, keep it open and take a scratch profile;
Firefox and Safari each want a driver protocol for that, and adopting one would
make the test runner own a browser automation library.

**With no browser, the run is refused, by name.** It does not fall back to the
shim: `--browser` was asked for because the shim's answer was not trusted, and a
green run on it would answer a question nobody asked. Five refusals, each
checked in `crates/uf_cli/tests/browser_host.rs` without needing a browser:

- no browser found, or `UF_BROWSER` naming something that is not there — the
  message lists every candidate it looked for;
- `--coverage` — `NODE_V8_COVERAGE` is Node's own switch, and the only way to
  ask a page for its profile is a protocol this mode does not speak;
- `-u` — a snapshot is a file beside the test that took it, and a page has no
  filesystem. Record on Node, check in the browser;
- a declared `permissions` set — it describes what a *process* may reach, and
  the process the tests run in is a browser uf started rather than a host uf
  configured. Sandboxing the driver instead would read as sandboxed in
  `uf explain` and leave the renderer the whole machine's network;
- no `node` on `PATH` for the driver.

Anything a page cannot have — `node:fs`, a synchronous `require` — throws a
sentence naming what is missing and what to do instead, at the moment it is
reached. None of them answers plausibly and moves on.

**One switch that looks like an exception and is not.**
`--remote-debugging-pipe` is passed and uf never writes a byte to it. Chrome
treats the pipe as its lifeline and shuts down when it closes, which is what
makes the browser die with the process that started it: `uf` kills a worker that
misses its deadline, a killed process runs no exit handler, and without the pipe
a timed-out file would leave a browser running with nobody to stop it. Measured:
SIGKILL the driver and browser processes go 11 → 0. A pipe rather than
`--remote-debugging-port`, which would open a port anything could speak to.

## The test that shows the difference

`tests/library/browser-layout.test.js` asks the same three questions on both
hosts and asserts the opposite answer in each, so neither host can change its
mind without the file going red. It passes in both modes:

```
uf test tests/library/browser-layout.test.js   →  4 passed   (happy-dom)
uf test --browser tests/library/browser-layout.test.js  →  4 passed   (Chrome 148)
```

Under the shim `getComputedStyle(half).width` is `"50%"`, every rectangle is
exactly zero and the `IntersectionObserver` never calls back; in a page the
width is `300px`, the rectangle is 300×40 and the observer fires. It reads which
host it is on from `typeof process === "undefined"` rather than from anything uf
sets — a test that asked uf which mode it was in would agree with uf even when
uf was wrong, and this file exists to disagree.

The assertions are two-sided rather than skipped on Node on purpose: the shim's
behaviour is half of what is being documented, and if happy-dom ever grows a
layout engine this should go red and be rewritten rather than quietly start
testing nothing.

## What it cannot do yet, written down

The issue calls this the smaller half of "Playwright-compatible browser
automation" and it is. **uf cannot drive this browser**: no clicking a real
cursor, no navigation, no screenshots, no throttling, no second tab. Those want
the protocol this mode deliberately does not speak, and VRT sits behind them
because a baseline is a screenshot. `fireEvent` and `userEvent` work exactly as
they do on Node, because dispatching a DOM event is a different thing from
driving a browser — and enough for a component test.

Also named in the docs rather than discovered: one browser process per worker
(so `-j 2` is worth typing on a small CI container), and that a file which
imports `node:fs` or spawns a process cannot run in a page. Point `--browser` at
the component tests, which is the half of a suite it is for.

`docs/roadmap.md` and `docs/hosts.md` are updated; the browser is deliberately
*not* a row in `uf_runtime::HOSTS`, because that matrix grades runtimes that run
a uf project and a browser already runs one — as the target runtime, which
`docs/red-lines.md` is careful to keep separate from the plugin runtime.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/728"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791537841&installation_model_id=422833&pr_number=728&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F728&signature=e85d573ab14e4e24c11453b9435f5c13d1e3eb5a6e114a3b9d17bf8b587be49d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Diagnosed the red CI here, since this has been sitting for an hour. It is **one**
failure, not three — `Upstream Flow Typecheck` runs the same suite and `CI` is
the gate reporting both:

```
---- shipped_modules_have_no_import_time_side_effects stdout ----
panicked at crates/uf_lib/tests/package_surface.rs:691:13:
test/browser-worker.js runs `main` at import time; a shipped module may only
declare, import and export at its top level
```

So `packages/test/browser-worker.js` calls `main` at its top level, and the
package-surface guard forbids that for anything `@uniflowed/*` ships.

The tension is real rather than accidental, which is why this needs your
judgement rather than a patch from me. A worker entry *is* a script that runs
when it loads — that is what makes it an entry and not a module — and the guard
exists because a shipped module that does work on import breaks tree-shaking,
makes import order significant, and runs before a caller can configure
anything. Both statements are true about this file.

Three ways out, roughly in order of how much I would trust them:

1. **Make the worker a module with an entry, not a module that is one.** Export
   the function and have whatever spawns the worker call it — a
   `self.onmessage` that dispatches, rather than a top-level `main()`. The guard
   then holds without an exception, which is the outcome that leaves no rule
   with a hole in it.
2. **Teach the guard about worker entries**, with the reason written into the
   test: an entry named in a `new Worker(...)` is a script by definition. That
   is defensible, but it needs the exception to be *narrow* — a path list would
   drift, so it wants a property the file has, not a name somebody typed.
3. Rename or relocate the file out of the shipped surface, if it turns out
   nothing imports it as a package export.

I have pushed nothing to this branch. `BEHIND` is not something you need to act
on — the merge queue rebases.

ubugeeei:
Merged `origin/main` in and fixed the guard failure — this had gone `DIRTY` as
well as red.

**The conflict** was in `docs/hosts.md`: both sides added a section at the same
anchor, yours ("The browser, which is a different question") and #744's ("What a
project may name, and which key decides"). Neither replaces the other, so both
are there.

**The failure** was the one I described earlier, and the answer turned out to be
already in the file. `ENTRY_POINT_MODULES` has carried `test/worker.js` since
before this PR, for exactly this reason — "modules that are process entry
points, and so run when they are loaded because that is what running them
means". Your two are the same thing:

* `test/browser-worker.js` reads `process.stdin` and calls `process.exit`.
* `test/internal/browser/page.js` takes over `console` when the page loads it —
  and says so itself: *"the same `internal/output.js` the Node worker
  installs"*.

So they join that list rather than getting a mechanism of their own, and the
comment above it now says what the three have in common: each is the beginning
of a run, not a module something imports for a name. I kept it a list of named
files rather than a directory or suffix, because a pattern would quietly cover
the next module written beside them, and the value of the rule is that a shipped
module doing work on import is a decision somebody made once, in writing.

That means none of the three options I offered earlier was needed — the second
one, in the shape this repository already had.

`cargo test -p uf_lib` green (52 across both binaries). No force-push, no
rebase.

Issue #206, opened 2026-09-05
ui: spreading rest props onto an intrinsic makes key mixed, which React rejects


ubugeeei:
Every part of `@uniflowed/ui` takes `...rest: { readonly [string]: mixed }` and spreads it onto a JSX intrinsic. Now that `uf check` resolves across modules (#199) and can see what the parts actually build, that spread is **35 of the package's 73 type errors**:

```
error[incompatible-type]: Cannot create button element because in property key:
  Either unknown [1] is incompatible with string [2].
  Or unknown [1] is incompatible with number [3].
    --> packages/ui/checkbox.js:55:6
     │
  55 │     <button
```

React's `key` is `string | number`; an indexer supplies `mixed`, so every intrinsic the package renders is rejected.

## What does not work

Declaring the property alongside the indexer:

```js
...rest: { readonly key?: string | number, readonly [string]: mixed }
```

changes nothing — measured, `packages/ui/checkbox.js` reports the same two errors with and without it. React's intrinsic checking reads the indexer regardless.

## What it probably needs

Rest props typed as React's own props for the element being rendered, rather than as an open indexer — the type `Rest` in `packages/ui/internal/merge-props.js` is the one place it is written down, even though the components spell it inline in 33 places. Whether Flow can express "the props of `button`, minus what this component controls" is the first thing to find out; if it cannot, the honest answer may be a named type per element rather than one `Rest`.

Worth doing as its own change: it touches every part of the package, and the current spelling is what `withoutComposed` and `mergeProps` are typed against.

The other 38 errors in the package are `React.Node` reading as `any`, which is the workspace-specifier half of #199 and not this.

PR #729, opened 2026-09-09
test(packages): move the JavaScript suite beside the code it tests

ubugeeei:
> **Merge after #722.** This branches off `cc/colocate-js-tests` and needs the
> `"!*.test.js"` it put at the end of all 49 manifests. Targeting `main` so the
> two land in order.

#722 did not move a file, and said why: `uf test#library` cannot see `packages/`,
and `shipped_modules()` would have held a co-located test to invariants written
for shipped modules. Both are fixed here, in that order, and then the 81 files
move.

---

## 1. The entry point moves to the repository root

A uf workspace member is any directory with its own `uf.config.js`
(`crates/uf_project/src/workspace.rs`), so `uf test#library` made `tests/library`
the project root — and discovery walks *down* from a root. `packages/` is a
sibling of `tests/`, so it was never walked. A file moved into `packages/ui` was
not a test that failed; it was a test that silently stopped existing.

`test:lib` now runs `./target/release/uf test` at the root. The `Library` CI job
already ran `uf run test:lib`, so it needed no edit — the door moved, not the job.

Measured on this branch point, before anything moved:

```
$ uf test#library      2851 passed, 1 skipped, 82 files
$ uf test              2851 passed, 1 skipped, 98 files, 39 unexpandable
```

**Same 2851.** The 16 extra files are shipped modules that mention
`describe`/`it` in a way discovery cannot expand — `packages/story/runner.js`
registers stories dynamically, `packages/validator/*.js` has the words in prose
— and they contribute no cases. That count is unchanged after the move, because
the files that moved were already in it.

### What running under the root config changes

Three things, named because two of them are invisible:

* **`ignore`.** `["upstream", "crates", "dist", "target", "node_modules"]` now
  applies to the suite, where `tests/library/uf.config.js` was `defineConfig({})`
  and applied the defaults. Nothing the suite needs is in that list; what it adds
  to the walk — `docs/`, `tests/type-tests/`, `tools/` — declares no tests, which
  is why the file count is 98 and not more.
* **`test.coverage.include: ["packages/"]` now includes the suite itself.** It
  does not count it, and the reason is a default rather than anything written in
  this repository: `CoverageConfig::exclude` defaults to `[".test.", ".spec."]`
  and `uf.config.js` sets only `include` and `reporters`, so the default stands.
  A test file is a project file the runner loads and not a file the report
  counts. Before the move `include` was what kept them out; now `exclude` is, and
  a future edit that widens `include` without knowing that would silently add
  eighty files at 100%. The comment in `uf.config.js` now says so.
* **`test:lib:coverage` no longer names a directory.** It was
  `uf test tests/library --coverage`; naming that directory now would measure the
  twelve files that stayed. It is `uf test --coverage`.

### `tests/library/uf.config.js` is deleted

Otherwise `uf test#library` would still work and would run whatever stayed
behind — twelve files, green, and nothing saying the other sixty-nine did not
run. That is the same failure this PR exists to remove, in a new place. Without
the config the selector is an error, which is the loud answer. `uf lint`,
`uf fmt --check` and `uf check` are root commands and never selected it.

## 2. `shipped_modules()` learns what a shipped module is

`package_surface.rs` defined it as every `.js` under `packages/`. A co-located
test would have violated three invariants immediately — its top-level
`describe(...)` is exactly the import-time side effect one of them forbids.

It now excludes test files, and **decides which those are from the manifest**,
not from a second list:

```rust
/// The `files` entry every shipped manifest ends with, which subtracts a test
/// file from whatever the entries before it added.
const TEST_FILE_NEGATION: &str = "!*.test.js";

fn is_test_file(path: &Utf8Path) -> bool {
    let suffix = TEST_FILE_NEGATION
        .strip_prefix("!*")
        .expect("the negation is `!` and `*` before the suffix it subtracts");
    path.file_name().is_some_and(|name| name.ends_with(suffix))
}
```

One string, read twice: `a_shipped_package_never_publishes_a_test_file` requires
it of a manifest, `is_test_file` derives "not shipped" from it. A file npm would
publish is held to every invariant below it; a file it would not is not a shipped
module. A package that stopped excluding test files fails that test rather than
quietly widening everything here.

`shipped_files()` is deliberately left whole — a test file is still a file in the
tree, so it is still a resolution target for
`every_relative_import_resolves_to_a_shipped_file` and still has to be a `.js` or
a `package.json` under `shipped_package_contains_only_modules_and_manifests`.

A new test keeps the exemption honest, and the second half is the one that would
rot: that `packages/` really does hold co-located tests, so the exclusion is
doing work rather than describing a case that never arises.

### The same fact, in the two other scanners over `packages/**/*.js`

Both would have been wrong in the quiet direction:

* **`tools/ci/publishable.sh`** files a package as a never-published
  *declaration* if any `.js` under it calls `nativeRuntimeRequired(`. A test
  asserting that refusal — "`withCalendar` raises `NativeRuntimeRequiredError`" —
  is one line, and it would have taken the package out of the release manifests'
  reach with no failure anywhere.
* **`published-packages.test.js`'s `sourcesUnder`** backs the `form → ui` edge
  rule. Reading a test file would report an edge no consumer can observe.

## 3. The files: 69 moved, 12 stayed

Mapped to the package whose behaviour a failure implicates, rather than to the
package with the most imports — most files import `@uniflowed/test` and a
fixture builder besides the thing they are about.

| package | files | | package | files |
|---|---|---|---|---|
| `router` | 14 | | `react-testing`, `hooks`, `web` | 2 each |
| `vite` | 10 | | `cell` `core` `effect` `fetch` `form` `graphql` `i18n` `immer` `mock` `query` `relay` `state` `std` `story` `stylex` `tui` `ui` `validator` | 1 each |
| `test` | 9 | | | |
| `server` | 8 | | | |
| `host` | 4 | | | |

**Twelve stayed, each because it is not one package's:**

| file | why |
|---|---|
| `deploy.test.js` | `server`, `vite` and the CLI's adapter output |
| `request-lifecycle.test.js` | the router/server seam — `after()` is the server's, the middleware runner is the router's, and the property is that they agree |
| `server-actions.test.js` | its own docblock names three: the bytes on the wire, the endpoint, and the tables the bundler builds |
| `assertion-types.test.js` | `@uniflowed/test`'s `expect` *and* `@uniflowed/react-testing`'s `fireEvent` |
| `docs-nav.test.js`, `docs-tables.test.js` | the docs site and `docs/security.md`, not a package |
| `lsp.test.js`, `mcp.test.js` | drive the `uf` binary over a wire; no package to sit beside |
| `vscode-extension.test.js` | `editors/vscode` |
| `published-packages.test.js` | every package at once, by `npm pack` |
| `in-source.test.js` | its subject is `in-source-subject.js`, a fixture. Beside `packages/test` that file would be an unexported, unpublished `.js` in a shipped tree — a violation of two invariants, to hold a fixture |
| `snapshot.test.js` | it carries `__snapshots__/snapshot.test.js.snap`, and `packages/` may hold a `.js` or a `package.json` and nothing else |

The last two are the interesting ones: both *do* test `@uniflowed/test`, and both
bring a non-module artefact with them. That is a rule about what may sit in a
published tree, not a filing preference — `shipped_package_contains_only_modules_and_manifests`
is the test that would fail.

`module-mock.test.js` did move, and its `.uf-module-mock-types/` fixture
directory moved with it, so `uf check` is now handed
`["packages/test/.uf-module-mock-types", "packages/test"]` — a path inside
another path it is given. Checked rather than assumed: 35 assertions, unchanged.
`.gitignore` follows.

## 4. Imports: the specifier for a surface, a relative path for an internal

One rule, and it falls out of what a test is for:

* **A public surface by its package specifier** — `@uniflowed/ui`,
  `@uniflowed/router/handler`. `exports` is part of the package's surface, and a
  test that reached past it into `./index.js` would pass while the export map was
  broken. Nothing changed here: 69 files kept every `@uniflowed/*` import they
  had, and Node resolves them from inside the package by self-reference and
  through the workspace symlink either way.
* **An internal by a relative path**, because `package.json#exports` deliberately
  does not name `internal/` — and now a short one:
  `./internal/routes.js` rather than `../../packages/vite/internal/routes.js`.
* **Another package's internals stay exactly as they were.** `tests/library/x.js`
  and `packages/vite/x.js` are both two directories deep, so
  `../../packages/react-testing/internal/dom.js` is the same string it was.
* **The suite's shared helpers stay in `tests/library/`** — `dom.js`,
  `type-tests.js`, `fixtures/` — reached as `../../tests/library/dom.js`. They
  serve every package, so putting them in one would be picking a package at
  random; putting them in a package at all would ship them.

## Verification

| check | result |
|---|---|
| `uf test` | **2852 passed, 0 failed, 1 skipped, 98 files** |
| `cargo test -p uf_lib` | 35 passed (34 on #722; +1) |
| `cargo test -p uf_check --lib` | 174 passed |
| `cargo clippy -p uf_lib --all-targets` | clean |
| `uf fmt --check` | every file already formatted |
| `uf lint` | 0 errors, 14 warnings — the documented steady state |

**2851 → 2852.** The one new case is `no package publishes a test file` in
`published-packages.test.js`, added below. No test moved and lost a case: the
count is identical either side of the move, and the delta is a test this PR
wrote.

### `npm pack --dry-run`, with real test files present

#722 proved the allowlist against probe files. These are the real ones — 26
packages, 69 test files on disk:

```
26 packages hold a co-located test; 69 test files on disk, 0 in a tarball

@uniflowed/router       14 on disk   17 packed  0 test files  [published]
@uniflowed/vite         10 on disk   21 packed  0 test files  [published]
@uniflowed/test          9 on disk   20 packed  0 test files  [published]
@uniflowed/server        8 on disk   28 packed  0 test files  [published]
@uniflowed/host          4 on disk    9 packed  0 test files  [published]
@uniflowed/ui            1 on disk   52 packed  0 test files  [published]
… 20 more, all 0
```

And the control, because "0 packed" is only worth something if something was
subtracted. With `"!*.test.js"` removed from the manifest:

```
@uniflowed/ui         with the negation: 0   without it: 1  (ui.test.js)
@uniflowed/hooks      with the negation: 0   without it: 2  (hooks-ssr.test.js, hooks.test.js)
@uniflowed/validator  with the negation: 0   without it: 1  (validator.test.js)
@uniflowed/vite       with the negation: 0   without it: 0
@uniflowed/test       with the negation: 0   without it: 0
@uniflowed/server     with the negation: 0   without it: 0
```

The last three are honest rather than impressive: their `files` are explicit
filenames plus `internal`, and every co-located test is at the package's top
level, so their allowlists would not have packed one anyway. The negation is what
makes that not depend on which shape a manifest happens to have — which is the
argument #722 made for putting it on all 49 rather than the 22 that needed it.

This is now a test rather than a transcript: `no package publishes a test file`
asks npm, and asserts a test file was there to exclude, so it cannot pass on a
repository whose suite moved back out.

## Every `grep -rn "tests/library"` hit

* `uf.config.js` — `test:lib`, `test:lib:coverage`, and three comments naming a
  test that moved.
* `.github/workflows/` — nothing to change: the `Library` job runs
  `uf run test:lib`, and the task is what moved.
* `tools/ci/` — `publishable.sh` (above), `security-scan.sh`'s comment.
  `docs-csp.sh` and `links-resolve.sh` name `deploy.test.js` and
  `docs-nav.test.js`, which stayed.
* `docs/` — 40-odd rows in `docs/security.md`, the guide and reference pages, and
  `architecture.md`. These are not cosmetic: `security:scan` fails a row that
  names a file which does not exist, so a stale path there is a red build.
* `editors/*/README.md`, `CONTRIBUTING.md`, `docs/issue-roadmap.md` — the
  `uf test#library` command form, now `uf test` or `uf test <path>`.
* `crates/` — sixteen doc comments naming a test, plus two Rust unit tests whose
  *input* was a path in `tests/library`. One of those needed thought:
  `any_of_several_patterns_selects` used `tests/library/ui.test.js` against
  `["packages/ui", "tests/"]`, and rewriting the path mechanically would have
  made it match the first pattern and stop testing the disjunction. It names a
  file that stayed.

## One thing the move surfaced

`uf check packages/ui` now reads `ui.test.js`, because `ui.test.js` is in
`packages/ui` — and reports
`Cannot create RadioGroup.Root element because in property key: unknown is
incompatible with empty` against the suite's own code, where `Plans` spreads
`Field.Control`'s render props onto a `RadioGroup.Root`.

`does not make React's key mixed` is an assertion about the elements the
*package* renders — #206 — so it now ignores diagnostics in test files, with a
second assertion that the checker read the package rather than only this file, so
the filter cannot empty the list it is asserting on. The finding is real and is
about `Field.Control`'s prop type; it belongs in an issue of its own, and it was
invisible until this file moved.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/729"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791537972&installation_model_id=422833&pr_number=729&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F729&signature=e9061c13f6de2a805a458f3e2a11eef808755bacd8fdebb2f9d617b78401883d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Diagnosed the red CI here, since this has been sitting for an hour. **The
failure is not in the move — it is a `uf_fmt` idempotence bug that the move
exposed**, so it needs fixing rather than working around.

`Test` fails at `./target/release/uf run rust:test`:

```
---- shipped_sources_are_formatted_idempotently stdout ----
panicked at crates/uf_fmt/tests/guarantees.rs:107:13:
assertion failed: `(left == right)`:
  packages/core/temporal.test.js is not idempotent
Differences (-left|+right):
                AFTERNOON + 185_500,
            )
```

`Upstream Flow Typecheck` fails on the same test through
`cargo test --workspace --all-features`, and `CI` is just the gate reporting
those two. So it is one failure, not three.

The construct is `packages/core/temporal.test.js:160`:

```js
    const later = Temporal.Instant.fromEpochMilliseconds(AFTERNOON + 185_500);
```

which is over the width, so the formatter breaks the argument out — and the
second pass over its own output does not reproduce it. `left` and `right` in
the assertion are pass 1 and pass 2 of the *same* file.

Why this PR surfaced it: `shipped_sources_are_formatted_idempotently` walks
`packages/`. The file was not under `packages/` before this move, so its content
had never been subject to the guarantee. Nothing about the move is wrong; the
guarantee simply reaches further now, which is an argument for the move rather
than against it.

Two ways forward, and the first is the one worth taking:

1. **Fix the formatter.** A call whose single argument is a binary expression,
   broken across lines, is not at its own fixed point. That is a real bug for
   every user, and this is the smallest reproduction anyone is likely to get.
2. Reflow that one line by hand so it is already at the fixed point. It turns CI
   green and leaves the bug in the formatter for the next file to find, so it is
   only worth doing if this PR is urgent and the fix is tracked separately.

I have not pushed anything to this branch. Say which you want, or if the fix
lands in `uf_fmt` first this goes green on a rebase — the merge queue rebases
anyway, so `BEHIND` is not something you need to act on.

ubugeeei:
Correction to my last comment, and a minimal reproduction.

**I named the wrong line.** The CI diff's `AFTERNOON + 185_500,` is the context
line *above* the change, not the construct. That call is stable on its own. The
unstable one is the `expect(...)` on the line after it, and specifically its
member chain.

Built a probe against `uf_fmt` directly. Under the narrow configuration the
guarantee test uses (`line_width` 40, `indent_width` 4, single quotes, no
semicolons) pass one gives

```js
        expect(
            at
                .until(later)
                .total({
                    unit: 'minute',
                }),
        ).toBeCloseTo(3.0917, 3)
```

and pass two, over that output, gives

```js
        expect(
            at.until(later).total({
                unit: 'minute',
            }),
        ).toBeCloseTo(3.0917, 3)
```

**It converges after one step** — pass 2 = pass 3 = pass 4. Only the first
transition is unstable, which is exactly what
`shipped_sources_are_formatted_idempotently` asserts against.

### What it takes to trigger

A ladder of six probes, all at width 40:

| case | stable |
| --- | --- |
| `expect(at.until(later).total({ unit: "minute" })).toBeCloseTo(3.0917, 3)` at top level | yes |
| the same, one arrow deep | yes |
| the same, **two arrows deep** (`describe` → `it`) | **no** |
| two arrows deep, chain but no object argument | yes |
| two arrows deep, object but no outer chain | yes |
| the chain alone, object written flat or already expanded | yes |

So it needs all three at once: enough indentation, an outer chain, and an object
argument. It is a width threshold, not a syntax the printer mishandles.

### The mechanism

`print_member_chain` ends in
`conditional_group([one_line, expanded])`, and a conditional group measures a
candidate only up to its first line break. Pass one's object is flat, so
`one_line` is flat, does not fit at that indent, and `expanded` wins — the chain
breaks. Pass two's object now carries a forced break, because uf preserves an
object's expansion when the author's source had a line break after `{`, and uf's
own output is now that source. `one_line` measures as `at.until(later).total({`,
which fits, so the chain stays flat.

The printer's output is the input to a rule about the input.

### The obvious fix is wrong, and I checked before proposing it

Refusing `one_line` whenever it will break makes all six probes stable and keeps
all 98 `uf_fmt` tests green — and reformats three committed files for the worse:

```js
-    permissions.query({ name }).then(
-      (result: PermissionStatus) => { … },
+    permissions
+      .query({ name })
+      .then(
+        (result: PermissionStatus) => { … },
```

`packages/hooks/browser.js`, `packages/i18n/negotiate.js` and
`packages/router/internal/runtime.js`. A chain whose last call takes a function
argument is *supposed* to stay flat and let the body break, so that condition is
far too broad. I reverted it rather than pushing it.

A real fix has to separate "the flat form breaks because a callback body breaks"
from "the flat form breaks because an object kept the author's newline", and
that is a change to the member-chain printer that wants its own PR and its own
argument.

### For this PR

Nothing here is your bug, and there is no reason to hold the move for it.
Reflowing that one line so pass one already lands on the fixed point unblocks
you, and I would take that — as long as the formatter bug is tracked rather than
absorbed. Say the word and I will push the reflow, or leave it to you if you
would rather the branch stayed yours.

ubugeeei:
Fixed properly — ubugeeei-prod/uf#746. Nothing for you to change here; once that
lands this branch goes green on the queue's rebase.

The narrow version turned out to exist. A chain whose flat form already contains
a forced break is not flat, so it is not offered as the flat candidate —
**unless the last call takes a function**, which is the exception that separates
the fix from the regression I showed you. With it:

* your `packages/core/temporal.test.js` is stable under all three of the
  guarantee's configurations;
* `permissions.query({ name }).then(cb)` still formats as one line plus a body;
* **0 of 302** shipped and templated files reformat — measured by formatting
  every one and diffing against what is committed.

`a_member_chain_is_its_own_fixed_point` covers both halves and fails without the
change with `an object argument is not idempotent`.

So the move stays as you wrote it, and the guarantee reaching further is the
argument for it rather than the cost of it.

ubugeeei:
Merged `origin/main` in — this went `DIRTY` when #740 and #741 landed. Merged,
not rebased, since others have merged into this branch.

Two conflicts, both the same shape: this branch moved the TUI suite and #741
changed what the comments beside it say, so each side was right about a
different half.

* `crates/uf_tui/src/lib.rs` — main's counts (selection landed, so ten claimed
  features and thirteen not-yet, not nine and fourteen) with your path,
  `packages/tui/tui.test.js`.
* `docs/issue-roadmap.md` — main's paragraph, which now describes selection as
  done rather than as still owed to #314, with the same path correction.

Git carried #741's seventeen new cases across the rename by itself: the moved
file has **118** cases, which is exactly the count that PR reported. So the move
did not cost the selection work anything.

`cargo test -p uf_tui` and `cargo test -p uf_lib` both green after the
resolution.

The formatter idempotence bug is fixed separately in ubugeeei-prod/uf#746, so
`temporal.test.js` needs nothing from you either.

ubugeeei:
Merged `main` in again — it now carries ubugeeei-prod/uf#746, the member-chain idempotence fix. That is what `shipped_sources_are_formatted_idempotently` was failing on once this branch moved `packages/core/temporal.test.js` under `packages/`, so the file needs no reflow and nothing in this branch had to change.

The move stands as you wrote it.

Issue #700, opened 2026-09-09
release: six packages are versioned every release and published never — `cell`, `state`, `effect`, `immer`, `motion`, `temporal`


ubugeeei:
Six packages are versioned and released with every other one and have never reached npm:

```
$ for p in cell state effect immer motion temporal; do
    printf '%-10s ' "$p"; curl -so /dev/null -w '%{http_code}\n' "https://registry.npmjs.org/@uniflowed%2F$p"
  done
cell       404
state      404
effect     404
immer      404
motion     404
temporal   404
```

Each has `"version": "0.0.0-alpha.18"` in its manifest and none is marked `private`, so the release bumps them and stops short of publishing.

`cell` is the one that sent me looking. A project already using a third-party reactive-cell library wants to move onto uf's, and the package it should move to exists — `packages/cell`, "Fine-grained reactivity: state, derived values, effects and resources" — but cannot be installed. `state` is in the same position, and #365 is a note-to-self in this repository that says so from the inside: "should be an atom in `@uniflowed/state` again, once that name is published."

Whether the fix is publishing them or marking them private, the current state is the one that costs a reader time: the source, the version and the description are all discoverable, and only `pnpm add` says no.

uf 0.0.0-alpha.18.

ubugeeei:
Checked the six against the repository's own classification, and they are not one group — they are two, with different fixes.

`tools/ci/publishable.sh` already refuses "in neither list", and it passes today:

```
$ sh tools/ci/publishable.sh
27 implemented packages: 15 published, 12 waiting on npm trust
```

So the split is:

| package | in `pending-packages.txt` | why it is not on npm |
| --- | --- | --- |
| `cell` | yes | waiting on `npm trust` — #210 |
| `state` | yes | waiting on `npm trust` — #210 |
| `effect` | yes | waiting on `npm trust` — #210 |
| `immer` | yes | waiting on `npm trust` — #210 |
| `motion` | **no** | classified declaration-only |
| `temporal` | **no** | classified declaration-only — #560 |

**Four of the six are #210 and nothing else.** They are implemented, listed, and blocked on the one step a checkout cannot take:

```
npm login
sh tools/release/bootstrap-publish.sh
sh tools/release/trust-npm.sh
```

`npm trust` can only bind a name the registry already has, so a name never published has to be published once by a person with a 2FA prompt. Nothing in this repository can move them.

[...]

Issue #560, opened 2026-09-07
release: `@uniflowed/temporal` is not publishable, so the package with the front door cannot be installed


ubugeeei:
`packages/temporal` is the front door for the Temporal surface, and it is in `tools/release/pending-packages.txt` rather than `published-packages.txt`. So the specifier that actually installs today is `@uniflowed/core/temporal`, and the package named after the feature is not on npm.

That is also why the implementation lives in `@uniflowed/core`: `publishable.sh` refuses a published package depending on an unpublished one, and `@uniflowed/hooks` is published. The layering is correct; what is missing is the last step.

**Done**: `@uniflowed/temporal` publishes, `@uniflowed/hooks` and `@uniflowed/web` can name it directly, and the docs stop having to explain that the package and the import path are different things.

Part of #509.

ubugeeei:
Waiting on the same two commands as ubugeeei-prod/uf#700, which now lists them and the thirteen names: `npm login && tools/release/bootstrap-publish.sh`, then `tools/release/trust-npm.sh`. Nothing else in the repository blocks this — binding a name is an account change npm gates behind a one-time password, which is why `trust-npm.sh`'s own header says no workflow and no agent can do it.

ubugeeei:
Reopening: #869 moved the source metadata and direct imports, but the issue's Done condition is not satisfied yet. The live registry still returns E404 for `@uniflowed/temporal`, and `uf@0.0.0-alpha.32` was tagged/published from 6d20b86 before #869 merged, so that publish run never sent temporal. The remaining work is still the npm-side first publish/trusted-publisher bootstrap, then a release that includes #869 and external registry/install verification.

ubugeeei:
Reopening again: the live npm registry still returns E404 for `@uniflowed/temporal` and `@uniflowed/temporal@0.0.0-alpha.33`. Release PR #871 is still draft/open, and no publish has completed for the temporal package. Keep this issue open until npm shows the package and install/registry verification passes.

ubugeeei:
[...]

Issue #210, opened 2026-09-05
release: four packages are implemented and not yet on npm


ubugeeei:
`uf@0.0.0-alpha.4` published all seventeen names in `tools/release/published-packages.txt` (#130). Four packages implemented since the last release are not in that file, because a name that `npm trust` has not bound fails the publish job **after** the names before it have already gone out — the file's own header says so, and adding them without binding first is how a release ends up half-sent.

- `@uniflowed/effect` — the effect system, measured against Effect-TS
- `@uniflowed/form` — the form store
- `@uniflowed/immer` — draft-based immutable updates
- `@uniflowed/mock` — MSW-compatible request mocking

## What has to happen, and why CI cannot do it

Both steps need an npm session and a one-time password, so they are ubugeeei's:

```bash
sh tools/release/bootstrap-publish.sh   # publishes only the names the registry lacks,
                                        # at the workspace version, on the alpha tag
sh tools/release/trust-npm.sh           # binds each name to publish.yml over OIDC
```

`bootstrap-publish.sh` shows its plan and asks before it does anything, and dry-runs every package before publishing any. `trust-npm.sh` writes npm's 2FA URL straight to the terminal — it used to capture the output and swallow it.

Then the four names go into `tools/release/published-packages.txt` and `tools/release/preflight.sh` will report twenty-one on the registry instead of seventeen.

Until then those packages exist in the repository and in the workspace, and not on npm. `verify-npm.sh --closure-only` is what would catch a published package depending on an unpublished one, and it passes today because nothing published depends on these four.

ubugeeei:
It is ten, not four, and the list is in the repository now rather than in this issue.

#229 added `tools/release/pending-packages.txt` and `tools/ci/publishable.sh`, which refuses a package that never calls `nativeRuntimeRequired` and is named in neither manifest — so the gap cannot widen again without CI saying so. The ten:

```
brand  cell  effect  form  immer  mock  state  story  testing  web
```

About 22,000 lines of Flow. `@uniflowed/state` is what this project offers instead of Jotai, `@uniflowed/effect` instead of Effect, `@uniflowed/form` instead of React Hook Form, `@uniflowed/immer` instead of Immer — and `npm install` answers `ETARGET` for every one of them.

## What is left, and only you can do it

[...]

PR #730, opened 2026-09-09
fix(release): a package that is never published says so in its manifest

ubugeeei:
Half of ubugeeei-prod/uf#700.

## The state it describes

Twenty directories under `packages/` are declaration modules — every exported function calls `nativeRuntimeRequired`, `tools/ci/publishable.sh` classifies them as such, and neither release manifest names them. No release has ever published one, and no release ever will.

Nothing said so. Each carries a `version`, `bump-version.sh` moves it every release, the description and the source are on GitHub, and the only thing that ever refuses is `npm install`, with `ETARGET` and no explanation. That is the state #700 says costs a reader time:

> the source, the version and the description are all discoverable, and only `pnpm add` says no.

`"private": true` is npm's own way of saying "not for the registry". Saying it in the manifest puts the answer where the reader already is.

## The rule, and why it is keyed the way it is

`publishable.sh` gains a third rule, for the reason its own header gives for the first two — *"a list somebody adds to is a list somebody forgets."*

**Membership of a release list wins over the scan, and it has to.** Two directories look like declarations to `isDeclaration` and are published on purpose:

| | why it matches | why it is right |
| --- | --- | --- |
| `core` | it **defines** `export function nativeRuntimeRequired(…)` | the regex cannot tell a definition from a call |
| `stylex` | three exports call it | they are **compile-time** — `uf transform` replaces the call with the object it computed, so reaching the body means uf was bypassed, and throwing is the correct answer |

Both are on npm today (`200` from the registry for each). A rule keyed on the scan alone would have marked them private and stopped a release that is correct. I very nearly shipped exactly that; the check is what caught it, which is the argument for keying on the list.

## Verification

- `sh tools/ci/publishable.sh` — passes: `27 implemented packages: 15 published, 12 waiting on npm trust`
- **Mutation-checked.** Removing `"private": true` from `packages/motion/package.json` makes it fail, naming the package:
  ```
  packages/motion is a declaration module in neither release list, so it is never
  published — but its manifest does not say so. Add `"private": true` …
  ```
- `uf test#library` — 2864 passed, 0 failed, 1 skipped
- `npm pack --dry-run` in `packages/ui` still packs its 52 files, so nothing about publishing a real package moved.

## What this does not do

The other half of #700 — `cell`, `state`, `effect` and `immer` are implemented, listed in `pending-packages.txt`, and blocked on the one step a checkout cannot take (`npm login`, `bootstrap-publish.sh`, `trust-npm.sh`). That is #210 and nothing here touches it.

It also does not decide `temporal`. It is private here because it is a declaration module **today**; #560 is the question of whether it should stop being one.

Refs ubugeeei-prod/uf#700

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/730"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791540709&installation_model_id=422833&pr_number=730&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F730&signature=9d02dfe9350f13f182d498ca45cbabc07468d3fcae2e7454f2010a265e06f24f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Release Management**
  * Marked internal packages as private to prevent accidental publication.
  * Added validation to ensure unpublished declaration packages are explicitly marked private.
* **Bug Fixes**
  * Updated release checks to allow packages listed as pending publication.
  * Added coverage for rejecting unmarked private packages and accepting pending packages.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/browser",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/browser, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/cli",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/cli, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
--- a/packages/hmr/package.json
+++ b/packages/hmr/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/hmr",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/hmr, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
--- a/packages/jsx-runtime/package.json
+++ b/packages/jsx-runtime/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/jsx-runtime",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/jsx-runtime, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/lib",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/lib, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
--- a/packages/lint/package.json
+++ b/packages/lint/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@uniflowed/lint",
   "version": "0.0.0-alpha.18",
+  "private": true,
   "description": "Flow declarations for @uniflowed/lint, part of the Unified Toolchain for Flow.",
   "type": "module",
   "license": "MIT",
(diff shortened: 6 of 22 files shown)


ubugeeei:
Pushed `35cc5c3d` on top of this branch — the `Metadata` failure was
`test-publishable.sh` not knowing about the rule this PR adds.

The fixture tree every scenario starts from is meant to be a *correct* one, and
its `packages/orm` had no `private`, so the new check refused it: `a described
tree was refused: … packages/orm is a declaration module in neither release
list`. Reproduced locally before touching anything.

Two things changed:

* the fixture marks `orm` private, because that is what a correct tree looks
  like once the rule holds;
* two scenarios cover the rule itself, which nothing did — a declaration in
  neither list with no `private` is refused *and named*, and a declaration a
  release list claims is not asked for one. The second is the half that keeps
  `core` and `stylex` publishable, since both look like declarations to
  `isDeclaration` and are published on purpose.

Verified the pair is not vacuous: with `manifest.private === true`
short-circuited in the check, `test-publishable.sh` fails with `a declaration in
neither list with no private was accepted`.

Locally: `tools/ci/test-publishable.sh` 14/14, `tools/ci/publishable.sh` exit 0
on the real tree (`27 implemented packages: 15 published, 12 waiting on npm
trust`), `tools/ci/scripts-parse.sh` 40 scripts parse.

No force-push and no rebase — this is a commit on top, so nothing you have is
lost if you were still working here.

PR #738, opened 2026-09-09
fix(nix, ci, docs): the toolchain the flake was missing, the build nothing ran, and the page that never had Nix on it

ubugeeei:
「CI Runner と Nix もファーストサポートね」 — the Nix half.

Nix was *present* here. It did not work, in three separate ways, and all three
were invisible for one reason: **nothing ever ran the flake.** The only Nix in
CI was `Formal` entering a dev shell to run `why3`, which never builds a crate,
so `nix build .#uf` had not been executed by anything since it was written.

## What was wrong

**1. The flake could not build, on any machine, on any day.** Both flakes named
`rust-bin.stable."1.98.0"`. `rust-toolchain.toml` pins `nightly-2026-08-01`, and
that is not a preference — 23 crates in Meta's Flow Rust port declare
`#![feature(box_patterns)]`, `uf_cli` reaches all of them through `uf_check`, and
an unstable feature attribute is a hard error on stable. `nix develop` also
handed you a shell in which the next line of `README.md` —
`cargo build --release --bin uf` — cannot succeed.

Fixed by reading the pin instead of copying it: `fromRustupToolchainFile
./rust-toolchain.toml`. There is now one pin, so it cannot drift from itself.

**2. The build skipped the upstream patches.** `postPatch` symlinked the pinned
Flow commit and stopped. `tools/upstream/sync.sh` defines a checkout as the
pinned commit *plus* `tools/upstream/patches/flow/`, "on every machine and in
every CI job". A build without them compiles, passes its tests, and answers
`uf check` differently from every other build of the same commit. It now copies
`rust_port/` and applies them — a symlink cannot be patched.

**3. Two flakes, and nothing said which.** `.envrc`, `formal.yml` and
`CONTRIBUTING.md` used `./tools/nix`; `README.md` said `nix develop`, the root.
They had already drifted (one had `nixpkgs-fmt`, the other OpenTofu).
`tools/nix/` is deleted; all four pointers name the root flake.

## What was missing

`overlays.default`, plus `nixosModules.default`, `darwinModules.default` and
`homeManagerModules.default` taking `programs.uf.enable` — nix-darwin included,
since that is what you run. The install page had **no Nix section at all** while
`README.md` said Nix was documented there; it now has one, beside the release
and from-source paths.

## What keeps it from rotting

- `.github/workflows/nix.yml` — evaluates all four systems on every change that
  touches the flake; asserts the dev shell's `rustc` *is* the pinned nightly
  rather than that the shell merely opens; builds `x86_64-linux` on `main`,
  weekly, and when the build's inputs change.
- `tools/ci/nix-first-class.sh` + `test-nix-first-class.sh` — the static half,
  no Nix needed, so `Metadata` runs it on every PR. Nine cases, each planted in
  a scratch repo and proven red.

## Verified by running, here

| Check | Result |
|---|---|
| `nix flake check --no-build --all-systems` | 43 outputs, all four systems, exit 0 |
| Dev shell toolchain | `nix derivation show` → `rust-minimal-1.99.0-nightly-2026-08-01`, no trace of 1.98.0 |
| Package toolchain | same nightly; `patch -p1 -d upstream/flow` present in the derivation |
| `nixpkgs-fmt --check flake.nix` | **caught my own flake unformatted**, fixed, re-checked clean |
| `tools/ci/nix-first-class.sh` | 17 assertions, exit 0 — it found the stale `./tools/nix` in `CONTRIBUTING.md` I had missed |
| `tools/ci/test-nix-first-class.sh` | 9/9 |
| `scripts-parse.sh` (dash), `gate-covers-every-job.sh`, `node --check uf.config.js` | exit 0 |

## Not verified here, and why

- **`nix build .#uf` was not run.** It compiles the whole workspace into the
  store (15 GB+). This Mac had 2–7 GB free with eleven agents building
  concurrently; I hit the 3 GB floor fetching a 47 MB formatter. So the claim
  "the package builds" rests on the toolchain and patch steps being *correct in
  the derivation*, which I did verify, and on `nix.yml` running the build — the
  first CI run of this PR is the first real build of this flake, ever.
- `uf run nix` as a task was not invoked (needs a built `uf`); the underlying
  script was run directly. `uf.config.js` is syntax-checked only.
- `links-resolve.sh` needs a built docs site and was not runnable.
- `cp -R ${flow}/rust_port` assumes that path exists in the pinned tree — it is
  what `sync.sh` checks out, but I did not fetch the input to confirm.

## Remaining

**Nothing keeps `flake.lock` current.** Dependabot has no Nix ecosystem, so the
`cargo` and `github-actions` entries leave it alone; the weekly `nix.yml` run
will surface staleness but cannot fix it. A scheduled `nix flake update` PR
(e.g. `DeterminateSystems/update-flake-lock`) is the follow-up — it needs write
permissions I did not want to add unasked.

`aarch64-darwin` is evaluated but never built in CI. It is the platform you'd
install from, and doubling a 30-minute job is the reason it is a documented gap
rather than a job.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/738"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791545363&installation_model_id=422833&pr_number=738&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F738&signature=0defb9c3267f1656eb9dfe6519eb543a1ddf72c94197b853b98e1c1a6eba9cf0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Pushed `974c5223`. Both red checks were real, and neither was the toolchain pin
— that part of this PR is right.

**`Build`**: the `postPatch` copied `rust_port` and nothing else, so it compiled
for twenty minutes and stopped at the first crate that embeds a library
definition:

```
error: couldn't read upstream/flow/rust_port/crates/flow_flowlib/src/../../../../lib/core.js:
       No such file or directory (os error 2)
```

`flow_flowlib` reaches *outside* `rust_port` with `include_str!`. Flow's own
definitions live in `lib/`, `prelude/` and `tslib/`, and the globals in
`evals/flow-typed/environment` are in none of those. `tools/upstream/sync.sh`
already names all five in `sparse_subtrees`, with a comment saying exactly this
would happen — so the flake now copies the same five, and
`nix-first-class.sh` holds the two lists equal by top-level name instead of
trusting a future edit to remember both places.

**`Dev shell`**: that check failed on a shell that was correct.

```
shell   uniflowed dev shell: Rust 1.99.0-nightly, Node v24.19.0, Bun 1.3.13
it read rust-toolchain.toml: a shell whose compiler cannot
```

The Rust is right — 1.99.0-nightly is the pin. The `shellHook` echoed its banner
to **stdout**, so `nix develop . --command rustc --version` returned the banner
instead of the version and the check read its own greeting back as a compiler.
A dev shell answers questions, and anything it prints on stdout corrupts every
answer, not just this one. Both echoes now go to stderr, with a guard that fails
on a `shellHook` echo that does not.

Both guards verified non-vacuous:

* copying only `rust_port` → `flake.nix copies [rust_port] and tools/upstream/sync.sh checks out [evals lib prelude rust_port tslib]`
* dropping one `>&2` → `flake.nix has 1 shellHook echo(es) on stdout`

Locally: `nix-first-class.sh` 19/19, `scripts-parse.sh` 42 scripts under dash,
`nixpkgs-fmt --check flake.nix` clean. I did not run `nix build` — that is the
half-hour compile this PR exists to make CI do, and CI is the first thing that
has ever done it.

No force-push and no rebase; this is a commit on top of yours.

ubugeeei:
Pushed `5bc7b06f`. The banner fix worked — the shell now answers the question:

```
shell   rustc 1.99.0-nightly (ad3d0bc14 2026-07-31)
```

and the check still failed, because **the check was wrong**. It compared
`rust-toolchain.toml`'s `nightly-2026-08-01` against that string, and a nightly
is built from the *previous* day's commit, so the date in `rustc --version` is a
day behind the channel date — always, for every nightly, on every machine. A
shell that is exactly right can never pass that assertion.

`rustc --print sysroot` is the toolchain's own directory and carries its name,
so the pin is in it verbatim. Confirmed against the pinned toolchain rather than
reasoned about:

```
$ rustc --version
rustc 1.99.0-nightly (ad3d0bc14 2026-07-31)
$ rustc --print sysroot
…/toolchains/nightly-2026-08-01-aarch64-apple-darwin
```

The version string is still printed — it is what a reader wants in the log — it
is just no longer what the assertion rests on.

That leaves `Build` as the only real question here, and it is the one this PR
exists to ask: the first `nix build .#uf` this flake has ever had.

ubugeeei:
Pushed again. **The build got past the compiler** — the subtree fix worked, and
`nix build .#uf` reached the end of `flow_flowlib` and every crate after it.
What failed is the `checkPhase`:

```
the JavaScript host is not available, so this test would prove nothing:
`node` is not on PATH; /build/source/node_modules/@uniflowed/vite/driver.js
does not exist; run `npm ci`
```

Twelve `tests/assets.rs` cases, all that message — which is that suite working
exactly as designed: a missing host *fails* rather than skipping, so nothing can
pass quietly for want of a fixture. A hermetic Nix build has neither `node` nor
an installed `node_modules`, and giving it both means an npm install inside a
build with no network. That is a much larger thing than this flake is for, and
it would duplicate what `ci.yml` already does on a machine that has all of it.

So `doCheck = false`, with the reasoning in the file rather than as a bare
disable. `nix flake check` still evaluates every output; `Test` and `Library`
still run the suite. This build asserts the one thing only it can: that the
package compiles from a pinned, patched, offline checkout — which is the claim
that had never once been tested before this PR.

`Metadata` was `test-nix-first-class.sh`, failing for the same reason
ubugeeei-prod/uf#730's did: its fixture is a repository with *nothing wrong*,
and two new rules made it wrong. The fixture now carries a miniature `sync.sh`
and a `shellHook` that greets on stderr, and two new cases cover the rules
themselves — a flake copying only `rust_port`, and a banner on stdout.

Locally: `nix-first-class.sh` 19/19, `test-nix-first-class.sh` 11/11,
`scripts-parse.sh` 42 under dash, `nixpkgs-fmt --check` clean.

Issue #403, opened 2026-09-06
check: `uf check` resolves no `@uniflowed/*` import, so this repository's own types check against nothing


ubugeeei:
`uf check` on any file in this repository ends with the same footer:

```
  › these imports are typed as any; uf resolved no module for them
    - @uniflowed/hooks
    - @uniflowed/react
    - @uniflowed/react-testing
    - @uniflowed/test
```

Every one of those is a workspace package with a symlink in `node_modules/@uniflowed/`, an `exports` map in its `package.json`, and Flow source behind it. `uf check` finds none of them, and the consequence is not a missing warning — it is that **nothing this repository writes against its own packages is checked at all**:

- `screen.getByRole("button")` is `any`, so `.valeu` is fine and `userEvent.click(anElement)` is fine even though `click` takes an `HTMLElement` and every query returns an `Element`.
- `import type { Control } from "@uniflowed/form"` is not a type import at all: `error[value-as-type]: Cannot use Control as a type because it is an any-typed value`. `tests/library/validator.test.js` has ten of these on `main` today, for `Schema`, `InferInput` and `InferOutput`.
- So the annotations `#225` just put into `tests/library/form.test.js` — `Control<{ title: string }>`, `Mode`, `ValidationRules`, `Resolver<Coded, Coded>`, `FormState` — are written from the package sources and are checked by nothing. They are right or wrong and the toolchain cannot say which.

Relative imports do resolve, but only when the checker is given the whole directory: `uf check packages/form/watch.js` reports `./internal/form-store.js` unresolved, and `uf check packages/form` resolves it. So the module graph is built from the file set rather than from the manifest, and a bare specifier is never followed.

What it needs, in the order they bite:

1. Follow a bare specifier through `node_modules` — including a symlink, which is what a workspace package is — and through the `exports` map, so `@uniflowed/form` reaches `packages/form/index.js`.
2. Load the module graph transitively rather than from the file set, so checking one file checks it against what it imports.

`#352` made a package resolve through the manifest that publishes it for `uf build`; this is the same question for the checker. Until it is answered, `uf check` cannot be added to `ci` for a reason much worse than a count of errors: it would be green about code it never read.

PR #739, opened 2026-09-09
fix(integrations, ci, docs): the install a `uf check` is worthless without, and the recipes read back against the binary

ubugeeei:
## What "supported" has to mean

A CI runner is *mentioned* when a page shows YAML somebody can copy. It is
**supported** when what we ship is executed against the CLI it claims to drive,
and when the facts a pipeline needs are stated by uf rather than guessed by the
reader.

`integrations/` already had the first half of that: three files that install uf
through the same checksum-verifying installer a human runs, and
`tools/ci/integrations-agree.sh` holding them to the variables the installer
reads. Nothing read the other half — what the pipeline does *after* `uf` is on
`PATH`. That was YAML for three systems this repository has no runner for, and
it was wrong.

## What was wrong, and why it was green

**`uf check` with no `node_modules` does not fail. It passes.** An import that
resolves to nothing is typed `any` — #403's design, and correct — so a check on
an uninstalled checkout reports "no problems" over a program it could not see.
Run here, against a fresh scaffold, with the binary from this tree:

```console
$ uf check                 # no node_modules
✓ no problems

  types checked  8
  › these imports are typed as any; uf resolved no module for them
    - @uniflowed/config
    - @uniflowed/react
    - @uniflowed/router
    - @uniflowed/test
```

`uf:check`, `uf:test` and `uf:build` in the GitLab template ran no `uf install`.
Neither did the CircleCI `uf/run` job, nor the orb's `install-in-your-own-job`
example. `uf test` and `uf build` fail there, which is honest. `uf check` went
green — the same failure mode the GitHub action refuses a Windows runner over,
shipped as the default recipe for two systems.

`uf fmt --check` is the quiet version. Its non-Flow formatter lives in
`node_modules/.bin`, and files it cannot format are skipped without reaching the
exit code. Run here:

```console
! biome is not installed, so 2 non-Flow files were skipped
! 1 file of 8 needs formatting         ← the Flow file, and only it
```

**The images could not have installed anything anyway.** `debian:stable-slim`
and `cimg/base` carry no Node and no package manager, so `uf install`,
`uf test` and `uf build` had nothing to run on. `uf` itself needs neither, which
is exactly why this was easy to miss.

**And the GitLab cache key named no architecture.** `uf-${UF_VERSION}`, for a
cache a whole fleet shares. uf publishes one archive per target; a fleet with
amd64 and arm64 runners restores one onto the other and reports a hit. The
action keys on `runner.os`/`runner.arch` and the orb on `{{ arch }}` — GitLab
had version alone.

## The change

| | |
| --- | --- |
| Install first | `uf install --frozen-lockfile` in every ready-made recipe, plus a `uf:fmt` job on GitLab and an `install-dependencies` parameter on the orb |
| Images | `node:24-bookworm-slim`, `cimg/node:22.11` |
| Cache keys | `uf-${UF_VERSION}-${CI_RUNNER_EXECUTABLE_ARCH}`, and a second GitLab entry for `node_modules` keyed on the lockfile with `key: files:` — different things answer to different keys |
| Verification | `UF_VERIFY_ORIGIN` reaches all three, as a `verify-origin` input/parameter. The installer has had keyless Sigstore verification and a second-opinion checksum host for a while and no integration exposed either; a pipeline is the one place that can install `cosign` and ask for `require` |
| Documentation | `/guide/ci` gains the install rule, exit codes and the machine-readable table, the two cache keys, the origin section, and the `jq` that fails a job when `uf check` resolved nothing |

## What checks it

**`tools/ci/recipes-are-runnable.sh`**, wired into `Metadata` as `ci:recipes`.

It reads every `uf` command out of the three integrations and out of
`/guide/ci`'s fenced blocks and asks the **binary** about each one — a shipped
recipe is a copy of the CLI's surface that ages on its own, and the binary is
the only source that cannot be stale. Five rules:

1. the subcommand exists;
2. every flag it passes exists;
3. its result can be read by a machine (`--json`), or it is named in
   `NO_MACHINE_READABLE_FORM` with the reason;
4. a command that reads `node_modules` is preceded by `uf install` **in the same
   recipe** — a unit is a top-level YAML key or one fenced block, so an install
   three jobs away is not this job's install;
5. the image can run what the recipe runs, and a toolchain cache key names the
   version *and* the architecture.

**Three of the five were failing before this change.** Verified by restoring the
three integrations to `HEAD` and running the gate — six findings, and then green
again once restored:

```
recipes: FAIL: integrations/gitlab/uf.gitlab-ci.yml:112 runs `uf check` with no `uf install` before it in `uf:check`
recipes: FAIL: integrations/gitlab/uf.gitlab-ci.yml:118 runs `uf test` …
recipes: FAIL: integrations/gitlab/uf.gitlab-ci.yml:124 runs `uf build` …
recipes: FAIL: integrations/circleci/orb.yml:206 runs `uf build` … in `examples`
recipes: FAIL: integrations/gitlab/uf.gitlab-ci.yml runs a command needing a JavaScript host on `debian:stable-slim`
recipes: FAIL: integrations/circleci/orb.yml runs a command needing a JavaScript host on `cimg/base:<< parameters.tag >>`
recipes: FAIL: integrations/gitlab/uf.gitlab-ci.yml caches the toolchain under a key naming no architecture
```

Two more were checked the same way and also fail as they should: a renamed
command (`uf check` → `uf typecheck`) and a flag uf does not have
(`uf fmt --dry-run`) — which is the staleness the whole file exists for.

**`tools/ci/test-recipes-are-runnable.sh`** is the check against a repository it
can safely make wrong, in the shape of `test-scripts-parse.sh`. Eleven cases:
one per rule, the two orderings that must not count (an install *after* the
command, an install in another job), that prose naming commands is not read as a
recipe, and that with no binary to ask it **fails** rather than reporting the
three rules it could still apply as a pass.

```
  ok  passes when the recipes are runnable
  ok  prose that names commands is not read as one
  ok  rejects a command that reads node_modules with nothing installed
  ok  rejects an install that comes after the command
  ok  rejects an install that belongs to another job
  ok  rejects a command the binary does not have
  ok  rejects a flag the command does not accept
  ok  rejects an image that cannot run what the recipe runs
  ok  rejects a toolchain cache key with no architecture
  ok  rejects a command a job would have to grep prose to read
  ok  fails rather than passing when there is no binary to ask
```

### `ci.yml`

**Two steps added to the existing `Metadata` job. No job added, removed or
reordered**, and no `needs:` touched, so the `CI` gate's list is unchanged.
`ci:gate` and `ci:gate:test` both pass (16/16 cases). The steps are
`./target/release/uf run ci:recipes` and `ci:recipes:test`, which is the form
every other step in that job takes; the tasks deliberately carry no
`dependsOn: ["build"]`, because `Metadata` downloads the `Toolchain` artefact
and has no Rust cache — a dependency on `build` would spend a release build
there to reproduce the binary already in the directory.

## Run versus read

**Run** (a real `uf` binary, on a scaffold outside this repository, because the
workspace makes every package resolve inside it):

* `uf check` and `uf check --json` with and without dependencies — the
  vacuous pass, and `typeCheck.untypedModules` as the field a job can fail on.
* `uf fmt --check` with an unformatted CSS, JSON and Flow file and no
  `node_modules` — the skipped non-Flow files and the exit code that ignores
  them.
* Non-interactive behaviour, with **stdin closed and `TERM=dumb`**: `uf` with no
  arguments prints the help on stderr and exits `2` rather than opening the
  menu; `uf nonsuch` exits `2`; `uf lint --json` exits `1` with pure JSON on
  stdout and the error line on stderr; `uf fmt --check` exits `0` clean and `1`
  dirty. Exit codes and the "no TTY" rule are correct today and needed no
  change — this is the part of the brief that turned out already to be true.
* `--json` coverage across the CI surface: `check`, `lint`, `test`, `inspect`,
  `explain`, `doc` have it. `uf fmt --check` and `uf build` do not.
* The new gate, its test, and every gate this change is near:
  `integrations-agree`, `install-banner-fits`, `gate-covers-every-job` and its
  test, `scripts-parse` (42 scripts) and its test. All green.
* `sh -n` under dash on both new scripts.

**Read**: the CircleCI and GitLab YAML dialects. I have no runner for either, so
`when:` steps, `key: files:`, `CI_RUNNER_EXECUTABLE_ARCH` and the `cimg/node`
and `node:*-slim` tags are read from their documentation and reviewed by eye —
no YAML parser is installed on this machine either, so the files were not even
parsed here. `.github/workflows/ci.yml` is parsed by GitHub and by `zizmor` on
this PR. The `Integrations` workflow, which is the one that installs a real
release on a real runner, is deliberately untouched: an assertion there about
`--json` fields would run against a *published* release rather than this branch
and could fail for reasons this PR is not about.

## What remains

* **`uf fmt --check --json`.** It is the one command in the CI surface a job has
  to read as prose, and it is a gap rather than a design — `check`, `lint` and
  `test` all have a document. It is a Rust change in `uf_cli` (`Fmt` gains
  `json`, `wants_json()` gains a case, and `fmt.rs` grows a payload beside its
  render), and I could not build it here: the shared cargo target directory has
  had its `deps/` removed, so any cargo invocation is a full rebuild of the
  graph, and the machine had 4 GB free with eleven agents on it. Until it
  exists, `NO_MACHINE_READABLE_FORM` in the new gate is where it is written
  down, so it is an exemption somebody chose rather than one nobody noticed. No
  cargo ran, so there is no `cargo fmt`/`clippy`/`cargo test` output to report;
  no crate was touched.
* **uf stating its own cache paths.** `/guide/ci` now says what to cache and
  what each key must include, and the gate holds the integrations to the
  architecture rule — but the source of those facts is a table in a document,
  not the binary. A `uf cache --json` (or a `caches` section in a `uf info
  --json`) would let a runner ask instead of copy, and would let rule 5 check
  the paths as well as the keys.
* **`uf build --json`.** Exempted here because the build manifest it writes is
  the machine-readable artefact, which is true but is not the same as a report.
* **Runners with no file here.** Jenkins, Buildkite, Woodpecker, Drone. The
  guide now ends with the four rules rather than an apology, because they are
  properties of uf rather than of the runner — but nothing checks a recipe this
  repository does not ship, and I have not claimed support for one.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/739"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791545832&installation_model_id=422833&pr_number=739&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F739&signature=20bcb57bf7565b5c89142adf8a7dc2edc0e3fec77fd0fbf6de712599d527e320"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -393,6 +393,16 @@ jobs:
       # and the failure surfaces as `uf: command not found` in a pipeline
       # belonging to somebody who cannot fix it.
       - run: ./target/release/uf run integrations
+      # And that the recipes inside those files still run. The step above
+      # checks the installation; this reads every `uf` command in the three
+      # integrations and in `/guide/ci` back against the binary this job
+      # downloaded, so a renamed command or a removed flag fails here rather
+      # than in somebody else's pipeline. It also holds the rule that costs the
+      # most to get wrong: a recipe must install the project before checking
+      # it, because `uf check` with no `node_modules` does not fail — it types
+      # every unresolved import as `any` and passes.
+      - run: ./target/release/uf run ci:recipes
+      - run: ./target/release/uf run ci:recipes:test
       # And that the set that goes to npm is closed under its own
       # dependencies. `@uniflowed/stylex` depends on `@uniflowed/core`, so a
       # release that sends one without the other publishes a package that
--- a/docs/app/_design/nav.js
+++ b/docs/app/_design/nav.js
@@ -199,7 +199,8 @@ export const sections: $ReadOnlyArray<Section> = [
       {
         href: "/guide/ci",
         title: "uf in CI",
-        blurb: "GitHub Actions, GitLab and CircleCI: one step each, and why they all pin.",
+        blurb:
+          "GitHub Actions, GitLab and CircleCI: one step each, why they all pin, and the install a check is worthless without.",
       },
     ],
   },
--- a/docs/app/guide/ci/_uf.page.mdx
+++ b/docs/app/guide/ci/_uf.page.mdx
@@ -1,6 +1,6 @@
 ---
 title: "uf in CI · uf"
-description: "GitHub Actions, GitLab and CircleCI: one step each, and why they all pin."
+description: "GitHub Actions, GitLab and CircleCI: one step each, why they all pin, and the install a check is worthless without."
 ---
 
 <p className="eyebrow">The tools</p>
@@ -41,13 +41,18 @@ jobs:
       - uses: ubugeeei-prod/uf/integrations/github-actions@uf@0.0.0-alpha.10
         with:
           version: 0.0.0-alpha.10
-      - run: uf install
+      - run: uf install --frozen-lockfile
+      - run: uf fmt --check
       - run: uf check
       - run: uf test
 ```
 
-Inputs: `version`, `cache`, `setup-url`, `repository`. Outputs: `version`,
-`bin-dir`, `cache-hit`.
+Inputs: `version`, `cache`, `setup-url`, `repository`, `verify-origin`, `token`.
+Outputs: `version`, `bin-dir`, `cache-hit`.
+
+The action installs the *toolchain*. `uf install` is a separate line because it
+is a separate thing — see [Install before you check](#install-before-you-check),
+which is the one mistake in this page that does not announce itself.
 
 On a Windows runner it **fails** rather than skipping. uf publishes macOS and
 Linux binaries only, and a matrix leg that quietly does nothing still reports
@@ -66,20 +71,23 @@ variables:
 check:
   extends: .uf
   script:
-    - uf install
+    - uf install --frozen-lockfile
     - uf check
 ```
 
-The template also ships `uf:check`, `uf:test` and `uf:build` as ready-made jobs,
-so a project with nothing unusual about it writes no `script` at all. They are
-ordinary jobs: extend, override or delete them.
+The template also ships `uf:check`, `uf:fmt`, `uf:test` and `uf:build` as
+ready-made jobs, so a project with nothing unusual about it writes no `script`
+at all. They are ordinary jobs: extend, override or delete them.
 
-Two details are GitLab's rather than uf's. The toolchain installs into
+Three details are GitLab's rather than uf's. The toolchain installs into
 `$CI_PROJECT_DIR/.uf-toolchain`, because GitLab's cache only carries paths
 inside the project directory — install it in `$HOME` and every job of every
-pipeline downloads the release again. And the default image is Debian rather
-than Alpine: uf ships a `*-unknown-linux-gnu` binary, and a glibc binary does
-not start on musl.
+pipeline downloads the release again. The default image is glibc rather than
+Alpine, because uf ships a `*-unknown-linux-gnu` binary and a glibc binary does
+not start on musl. And it is `node:24-bookworm-slim` rather than a bare Debian,
+because `uf install` drives the project's own package manager and `uf test` and
+`uf build` run on a JavaScript host: `uf` itself needs nothing, and three of the
+four things a pipeline asks it to do need Node on the image.
 
 ## CircleCI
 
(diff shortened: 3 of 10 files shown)


ubugeeei:
Two things this repository's own checks caught after the first push, both folded
into the commit. Worth writing down because they are the same argument the PR
makes, applied to me.

**`Flow format`.** `uf.config.js` is a project file, so `uf fmt --check` reads
it, and the `inputs` array I added was wrapped where the formatter would have
put it on one line. Nothing local told me: `uf fmt --check` here skips the
non-Flow half for want of a Biome, and I read the exit code rather than running
the whole project's check. Fixed, and `uf fmt --check` is clean on both JS files
this branch touches.

**`Library`.** `docs/app/_design/nav.js` blurbs each page with that page's own
`description` frontmatter, and `tests/library/docs-nav.test.js` asserts the two
are the same sentence. I rewrote the CI page's `description` and left the nav
saying the old one. It failed three times running and I spent two of those
looking at the wrong thing — the run's summary named a phantom test called
"a test nobody wrote", which is a line `console-output.test.js` *prints on
purpose* to prove a test cannot forge a protocol event; the real failure was one
`✗` line above it in a 3,800-line log.

Both are cases of a fact stated in two places with one of them wrong, which is
what `recipes-are-runnable.sh` is for — and neither would have been caught by
reading the diff. Green now, including `Metadata`, where the new gate runs, and
all four `setup-uf` legs, which install the changed action on real ubuntu and
macOS runners.

Issue #718, opened 2026-09-09
rsc: `node_modules` is ignored, so a `"use client"` inside an installed package is invisible — including `@uniflowed/ui`'s 43


ubugeeei:
Two facts in this repository, and the consequence that falls out of them.

**One.** `crates/uf_rsc/src/project.rs:27-28`:

```rust
pub const IGNORED_DIRECTORIES: &[&str] =
    &[".git", ".uf", "coverage", "dist", "node_modules", "target"];
```

The RSC scan skips `node_modules` outright.

**Two.** `packages/ui` ships `"use client"` in **43 modules**:

```
$ grep -rln '"use client"' packages/ --include='*.js' | grep -v node_modules | wc -l
43
```

`switch.js`, `alert-dialog.js`, `collapsible.js`, `combobox.js`, `date-picker.js`, `toggle.js`, `tooltip.js`, `field.js`, `tabs.js`, `resizable.js` and the rest.

**So in a user's application, where `@uniflowed/ui` is installed under `node_modules`, uf's own analysis never sees a single one of them.** The directives are correct, shipped, and unreadable by the tool that exists to read them — and the manifest they should feed is the one driving every substitution in `packages/vite/internal/rsc.js`.

Inside this repository the same modules are workspace members rather than dependencies, so the scan does reach them. That is what makes it easy to miss: uf's analysis is right about uf's own tree and blind about every tree that installs uf.

## Why ignoring `node_modules` is nonetheless the right default

Walking a dependency tree for directives is expensive and mostly pointless — almost nothing in `node_modules` is part of the application graph, which is exactly what `IGNORED_DIRECTORIES` is for. The fix is not to delete the entry. It is that **a package has to be able to declare a client root**, so uf can learn "`@uniflowed/ui/switch` is a client module" without walking to find out. React's own ecosystem answers this with the `"react-server"` export condition; a manifest field or an export condition would both do.

## Where it bites

- Any application using `@uniflowed/ui` with server components: a `<Switch>` in a server tree cannot be recognised as a boundary, so it is neither split nor reported.
- It is a prerequisite for #519. Found while working on that: `@uniflowed/router`'s `Link` has the same problem, and the `rsc-split-app` fixture's own root layout imports it — so the fixture's layout cannot become a server component until a package can say where its client edge is.
- `rsc/server-only-import-in-client` and the rest of the RSC lint rules are decided from the same graph, so they are quiet about a package for the same reason.

## Not claimed here

[...]

Issue #519, opened 2026-09-07
rsc: streaming, and the payload the client re-renders a tree from


ubugeeei:
`docs/roadmap.md`: "Splitting a *module* — dropping a Server Component that sits above a boundary — is not, and is blocked on a Flight-shaped payload the client can re-render a tree from."

That payload is this issue. Route-level splitting works today; what does not is a server component tree arriving as a stream the client can apply, which is what makes Suspense boundaries resolve independently rather than the document waiting for the slowest one.

Related: #252 — the analysis is complete and nothing reads it.

**Done**: a Flight-shaped payload, a client that applies it, a boundary that resolves on its own, and a test that asserts the *order* bytes arrive in rather than only the final document — because a stream that produces the right document at the end and streams nothing is the failure this is about.

Part of #488.

ubugeeei:
Sizing this, because #636 needed the answer and it is not what the issue implies.

**Streaming is not the missing half.** `packages/router/internal/stream.js` has run through `renderToPipeableStream` and `renderToReadableStream` for a while now — `renderDocument` picks per host, `prerenderDocument` uses `react-dom/static`, and there is no live `renderToString` call site left in the tree. Suspense boundaries already resolve independently in the HTML stream. What is missing is the payload a **client** re-renders a tree from.

**And that payload is a second module graph, not a second dependency.** Checked rather than assumed, against the pinned `react` 19.2.8:

- `react-server-dom-webpack/server.node` throws on import unless the `react-server` export condition is on. Verified: `node probe.mjs` fails, `node --conditions react-server probe.mjs` gives you `renderToPipeableStream`, `registerClientReference`, `decodeReply` and the rest.
- The condition resolves `react` to `react.react-server.js` — the build with no `useState` in it. `react-dom/server`, which turns the payload into HTML, needs the ordinary one. Two builds of React cannot share one module registry.
- So a Flight renderer needs its own graph. Three shapes exist: `node --conditions react-server` for the whole process (breaks `react-dom/server` in it); a worker thread with those `execArgv` (verified to work — the worker resolves the server renderer while the main thread keeps `react-dom/server`); or a bundler environment resolved with that condition.
[...]

PR #742, opened 2026-09-09
feat(router, docs): the rows a payload arrives in, and the boundary each one resolves

ubugeeei:
## What

#519 asks for four things: a Flight-shaped payload, a client that applies it, a
boundary that resolves on its own, and a test that asserts the *order* bytes
arrive in. This is all four, for the payload's **data** half. The element half —
a client re-rendering a *tree* — is not here, and the last section says exactly
why and what is left.

A document uf streams carried one thing the browser read back: the loader's
answer, as one `JSON.stringify` of one finished value in `<script
id="__uf_data">`. So a page whose data was one slow thing and four fast ones
said nothing about the four until the one arrived, and a promise anywhere in
that value became `{}` — silently, which is the worse half.

`packages/router/internal/payload.js` makes it a payload rather than a value:

```
0  {"post":{"title":"…"},"comments":"$P1","related":"$P2"}
2  {"value":[{"id":7}]}
1  {"value":[{"body":"…"}]}
```

Row 0 is the model, with a `"$P<n>"` reference wherever the loader left a
promise. Each later row is one of those values, in a `<script
data-uf-row="n">` React streams into the *same response* when it settles — each
inside its own `<Suspense>`, so two of them arrive in the order they resolved
and not in the order they were written.

`packages/router/internal/payload-rows.js` is the browser's half: it reads the
rows already in the document, watches for the rest, and hands each one to the
promise the model referred to. So a boundary the server completes *after*
`hydrateRoot` resolves with no second request.

The API is `<Suspense>` and nothing else — no `defer()` to call, no `<Await>` to
render, because a promise in the loader's answer already says what either of
them would:

```js
export async function loader({ params }: LoaderArgs) {
  return {
    post: await readPost(params.slug),     // waited for
    comments: readComments(params.slug),   // not waited for
  };
}
```

## Why only one reference kind

`internal/action-wire.js` already said, under "What is deliberately absent",
that a decoder which reconstructs Flight's references "is a decoder that
constructs attacker-chosen objects", and that its grammar "is not the place to
grow one quietly". This is the place, and it grows exactly one: `"$P<n>"` names
a **row of this same payload**, resolved by a promise this module made.
Nothing in a payload names a constructor, a module, an export or a class, and
no string in one is looked up in any registry.

The two references the element payload needs — a client module and an element —
are precisely the ones a decoder resolves by *calling* something the payload
named, and those are what wait for the second React module graph. `$` is a
closed namespace: an unrecognised tag is **refused** rather than passed
through, so adding one later is a change to a list rather than to a decoder
that had already been letting it through.

## Why nothing else changed

Everything else in a payload is what `JSON.stringify` carries, unchanged. A
`Date`, a class with a `toJSON`, an `undefined` property, a `NaN` — each
crosses exactly as it did. Narrowing the loader's answer to the action wire's
closed grammar would be a contract change this issue is not about, made in a
throw, inside a loader, on somebody whose page worked yesterday.

Both walks hand back the value they were given when there is no promise and no
`$` string anywhere in it, so:

* a page with ordinary data produces the bytes it produced before this existed
  (asserted), and
* the browser hands the application the object `JSON.parse` made rather than a
  copy of it (asserted by identity, not equality).

The one hazard the rebuild would have introduced and `JSON.parse` does not —
`target[key] = value` for `__proto__` calling the setter on `Object.prototype`
— is closed with `defineProperty`, and asserted.

## A streaming bug found on the way, which is not the payload's

React cannot flush a segment while that segment holds an unresolved boundary,
and every component between uf's render root and a route — `RenderProvider`,
`RouterProvider`, `RouteView`, the error boundaries — renders no element of its
own. So a `<Suspense>` at the top of uf's tree is a direct child of the **root**
segment, and a document with one streams nothing at all.

Measured against the pinned React 19.2.8, four shapes, same promise, same
120 ms:

| tree | first byte |
| --- | --- |
| `[<div>, <Suspense>]` | 122 ms |
| `[<div>, <Suspense>, <div>]` | 122 ms |
| `<Suspense>` alone | 121 ms |
| `[<div>, <span><Suspense></span>]` | 2 ms |
| `<div><p/><Suspense></div>` | 1 ms |

Any host element between the root and the boundary fixes it. The payload's rows
are inside a `<span hidden>` for that reason, and removing it fails four of the
tests here.

**A route whose `_uf.loading.js` sits above no layout is in exactly the same
position and is still affected** — `RouteView` puts that boundary at the same
depth, and such a route does not stream today. That is not this change's bug and
this change does not fix it: the fix is a host element at uf's render root,
which is markup every document would grow, and that is a decision rather than a
patch. It is written down in `docs/architecture.md` and reported on #519.

## Tests, and that they fail without this

`tests/library/payload.test.js`, 24 tests in four sections: the format, the
document reader, the order the bytes arrive in, and the browser applying them.

The streaming assertions are made on chunks and **without a clock**. The first
chunk is read while neither promise has settled and nothing is scheduled to
settle them, so a renderer that waited for the data would still be waiting at
the assertion — no timer coming to rescue it. (`streaming.test.js` uses a
120 ms deadline for the same claim; that is the more direct spelling and the one
that flakes on a machine running eight test files at once, which it did here
before this was rewritten.)

Verified by reverting rather than by assertion:

* `git stash` of the two wired files (`internal/runtime.js`, `client.js`), with
  the new modules left in place → **5 failed**, and they are the five
  behavioural ones: all three streaming tests and both browser tests. The pure
  format tests still pass, because they drive modules that would not exist at
  all without this change.
* removing only the `<span hidden>` → **4 failed**: nothing streams, and the
  three tests that read a first chunk time out.

## Checks

| command | result |
| --- | --- |
| `cargo fmt --all` | clean, no Rust file changed |
| `cargo clippy -p uf_rsc --all-targets -- -D warnings` | `Finished dev profile in 4m 53s`, exit 0 |
| `cargo test -p uf_rsc` | `227 passed; 0 failed` plus `1 passed` doc-test, exit 0 |
| `uf test#library` (2888 tests, 83 files) | `✓ 2888 passed, 0 failed, 1 skipped` |
| `uf lint` | 0 errors, 14 warnings (all pre-existing) |
| `uf fmt --check` | every file already formatted |
| `tools/docs/build.sh` | build succeeded |
| `tools/ci/links-resolve.sh` | 2480 links, ok |

No Rust changed, so `uf_rsc` is unaffected by this branch; it is reported
because the issue asked for it.

One thing the local run did not cover and CI did: `uf_lib`'s
`the_shipped_surface_uses_modern_flow_spellings` refuses the deprecated `+prop`
variance sigil in a published package, and `payload-rows.js`'s two `interface`
types were written with it. Fixed in the two commits after the first, and
`cargo test -p uf_lib --test package_surface` is green locally (36 passed).
The indexer form `{ +[string]: … }` the payload itself uses is a different
thing the rule does not touch, and ten shipped modules already spell it that
way.

## What is left of #519, and why

Not done, and each for a reason rather than for time:

1. **The element payload** — a client re-rendering a *tree*. It needs
   `react-server-dom-*`, which loads only under the `react-server` export
   condition, which resolves `react` to the build with no `useState` in it while
   `react-dom/server` needs the ordinary one. That is a second module graph, so
   `@uniflowed/vite` grows an environment first. Unchanged from what
   `docs/architecture.md` and this issue's own comments already record.
2. **A client-module reference in the payload** — `$` has room for the tag and
   deliberately no meaning for it yet. Naming a client boundary by chunk id
   versus by specifier is an open contract decision on #718, and this change
   does not settle it: `RscManifest`'s boundary shape is untouched.
3. **The router redesign** the element half needs: `ResolvedRoute` split by
   serializability, `RouteView` rendering a node it was handed, and
   `@uniflowed/router`'s export map split so a root layout can be a server
   component. This change touches none of it — the payload carries data, and
   data has no modules in it.
4. **A `<Suspense>` at uf's render root still does not stream**, above.

Refs #519

### DIFF
--- a/docs/app/guide/routing/_uf.page.mdx
+++ b/docs/app/guide/routing/_uf.page.mdx
@@ -306,6 +306,61 @@ export default component Post(data: {| +post: Post |}) {
 rather than returned, so the type of the data does not have to admit "or a
 redirect".
 
+### Deferring part of the answer
+
+A loader may leave a promise in what it returns. The page is rendered with the
+promise still pending, and the value arrives in the same response, later:
+
+```js
+export async function loader({ params }: LoaderArgs) {
+  return {
+    post: await readPost(params.slug),      // waited for
+    comments: readComments(params.slug),    // not waited for
+  };
+}
+
+export default component Post(data: {| +post: Post, +comments: Promise<Array<Comment>> |}) {
+  return (
+    <article>
+      <h1>{data.post.title}</h1>
+      <Suspense fallback={<p>Loading comments…</p>}>
+        <Comments comments={data.comments} />
+      </Suspense>
+    </article>
+  );
+}
+
+component Comments(comments: Promise<Array<Comment>>) {
+  return <ul>{use(comments).map((one) => <li key={one.id}>{one.body}</li>)}</ul>;
+}
+```
+
+The document the server writes carries the post, the heading and the fallback
+immediately; when `readComments` settles, the comments' markup **and** the data
+behind them are appended to the same response, and the browser resolves that
+boundary without a second request. Each deferred value is its own boundary, so
+three of them arrive in the order they resolved rather than in the order they
+were written.
+
+This is uf's payload format, and `<Suspense>` is the whole of the API: there is
+no `defer()` to call and no `<Await>` to render, because a promise in the data
+already says everything either of them would have.
+
+Three things are worth knowing about it:
+
+- **What a promise resolves to is data.** The same values a loader could always
+  return — JSON's — and not a promise of its own. A deferred value that is
+  itself deferred is refused, because the payload numbers its rows once.
+- **A rejected promise reaches the boundary that was reading it.** The browser
+  is told the value failed and not why; the reason stays on the server, where
+  it is logged, and reaches the error boundary above whatever used the value.
+- **The response is already a 200.** The same rule as a deferred loader: bytes
+  that have gone cannot be unsent, so a `notFound()` thrown from inside one of
+  these arrives at the error boundary rather than at the 404 page.
+
+`uf build` waits for every one of them: a file on disk has nothing to stream to,
+so a prerendered page contains the finished comments.
+
 ## Metadata
 
 A page or layout may export a `metadata` object, or a `generateMetadata`
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -701,9 +701,42 @@ process, a worker thread started with those `execArgv`, or a bundler
 environment resolved with that condition — which is the shape `@uniflowed/vite`
 would have to grow, since the first two are Node-only and uf's edge, serverless
 and workerd adapters all serve the same `handler.js`. That is the size of
-ubugeeei-prod/uf#519, and it is why the answer is not a smaller version of
-itself: half a payload format in the tree is the worst state for the thing
-whose whole risk is deserialisation.
+ubugeeei-prod/uf#519, and it is why the *element* half of the answer is not a
+smaller version of itself: half of it in the tree is the worst state for the
+thing whose whole risk is deserialisation.
+
+The framing is separable from that, and it has landed.
+`packages/router/internal/payload.js` is the wire format: a payload is a
+sequence of numbered rows rather than one value, row 0 is the model with each
+unresolved value replaced by a `"$P<n>"` reference, and each later row is a
+`<script type="application/json" data-uf-row="n">` React streams into the
+document at the moment that value settles. A loader may therefore leave a
+promise in what it returns — the shell goes out with a fallback and the value
+follows in the same response — and rows arrive in the order they resolved
+rather than the order they were written.
+
+It carries **one** reference kind, and that is the whole of why it could land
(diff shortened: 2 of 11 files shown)


ubugeeei:
Pushed a commit on top — the only red check was
`the_shipped_surface_uses_modern_flow_spellings`:

```
4 deprecated Flow spellings in the shipped surface:
router/internal/payload-rows.js:83: `+prop` is now `readonly prop`
router/internal/payload-rows.js:84: `+prop` is now `readonly prop`
router/internal/payload-rows.js:89: `+prop` is now `readonly prop`
router/internal/payload-rows.js:90: `+prop` is now `readonly prop`
```

The four are the properties of `DocumentLike` and `ElementLike`. `+x: T` and
`readonly x: T` are the same property, so this is spelling and not behaviour —
but it is spelling in the surface uf ships, which is what that guard exists to
keep current.

`cargo test -p uf_lib --test package_surface` now passes 36/36. `Test` and
`Upstream Flow Typecheck` were both reporting this one failure, and `CI` was
the gate reporting them, so that should be all three.

No force-push and no rebase.

Issue #947, opened 2026-09-14
ui: `uf ui add` — styled components a project owns, added by one command, the way shadcn/ui works


ubugeeei:
`@uniflowed/ui` is headless — keyboard maps, ARIA contracts, composition types, no styles — which is React Aria's half of the job. The half people adopt shadcn/ui for is different: a styled component **in your own repository** that you read and change, added by one command. uf has nothing for that half.

## What to build

- `uf ui add <component…>` writes a styled Flow component into the project (`app/components/ui/<name>.js` unless configured otherwise), built on the `@uniflowed/ui` primitive and styled with StyleX against the tokens `@uniflowed/stylex` compiles, and adds any dependency it needs through the project's package manager.
- `uf ui list` shows what the registry has and what the project has installed. `uf ui diff <component>` shows how the project's copy differs from the registry's current version — the answer to "how do I update a component I have edited".
- The registry is data in this repository, versioned with uf: one directory per component with its source, dependencies, an example and its accessibility notes.
- The components reference renders every registry component live, beside its source.

## Done when

- [ ] A scaffolded app runs `uf ui add button dialog tabs select`, builds, and passes `toHaveNoAxeViolations` on a page using all four.
- [ ] Every component `@uniflowed/ui` ships has a registry entry.
- [ ] `guide/ui` explains owned versus headless components and when to reach for each.

ubugeeei:
Progress on `uf ui add`.

**Merged**

- #995: uf's StyleX compiler accepts a condition that selects the state a headless part announces, e.g. `:is([aria-selected=true])` or `:is([data-active=true])`. It sorts after `:hover` and the form states, so a pressed toggle stays pressed under the pointer. The styled layer needs this to draw a select's keyboard cursor, which no pseudo-class can reach because focus stays on the trigger.

**Open**

- #1025: the registry (`registry/ui/`, embedded into the binary by `crates/uf_ui`), `uf ui add`, `uf ui list` and `uf ui diff`, and the first four components (`button`, `dialog`, `tabs`, `select`). The PR body lists the decisions this issue asked to be written down: where the registry lives, the per-component format, where files go and how imports resolve, theming and dark mode, how a copy records the version it came from, and what happens to a file that already exists. `crates/uf_ui/src/lib.rs` argues each at length.

[...]

PR #1109, opened 2026-09-15
fix(vite): name a module's StyleX stylesheet by its path from the project root

ubugeeei:
## What changed

Each module that compiles `stylex.create` imports a stylesheet of its own, named `uf-style:<id>.css`. The id was the module's absolute path. In a client chunk that virtual module is one of the chunk's sources, so the published source map of any client component styled with StyleX carried the absolute path of the machine that built it.

`security:scan` found this on #1098: `/home/runner/_work/uf/uf` in the components reference's client chunk map, which is the first client StyleX code the documentation ships. The finding reads:

```text
"sources":["../../../uf-style:/home/runner/_work/uf/uf/registry/ui/accordion.js.css", …
```

The stylesheet is now named with `moduleId`, the root-relative spelling the module graph report already uses for the same reason. An absolute path makes two machines' builds of the same project disagree, and publishes where one of them keeps its files. A module outside the root, such as a workspace package, climbs out with `../`. The name only has to be unique and stable, and the root-relative path is both. Every other use of the id matches on the `uf-style:` prefix and is unchanged.

## Verification

- `uf test` over the plugin's module-graph, flight, rsc-split and document-assets tests: 39 passed. `uf fmt --check` and `uf lint` on `packages/vite/index.js`: clean.
- `moduleId`'s own tests already cover a root-relative id and a module outside the root.
- A docs build carrying both #1098 and this change, then `tools/ci/security-scan.sh`: result in a comment below.

This unblocks #1098. Part of #947.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #1110, opened 2026-09-15
fix(router): write the Flight payload only between elements

ubugeeei:
## What changed

`interleaved` in `packages/router/internal/stream.js` wrote each waiting Flight payload chunk right after the HTML chunk it had just written. React hands its HTML on through a fixed-size buffer, so that chunk can end inside a tag, an attribute value, a comment or a character reference. A payload element written there is not an element. It becomes part of the attribute, comment or text, the browser never reads it, and React's client closes the payload with rows missing. The page reports "Connection closed" instead of hydrating.

On main it affects every page of the documentation. For each built page, `flight-placement` counted its payload's data element inside markup, and only the end marker between elements. On most pages the data element sits inside the site navigation:

```html
<li><a href="/gui<script type="application/json" data-uf-flight>"1:\"/assets/client-reference-3-….js\"\n2:…
```

`docs:links` caught it on #1098, as links to `/gui<script type=` and `/refer<script type=`.

A waiting payload is now written only where the HTML written so far ends with a `>` outside a `<script>` or a `<style>`. `advanced` tracks that across chunks and carries a tag split between two chunks into the next one. React escapes `<` and `>` in text and in attribute values, so a `>` it wrote closes a tag or a comment. The writer's other rules are unchanged: nothing before the head, and `</body></html>` after the end marker.

## Verification

- A new case in `packages/router/flight-stream.test.js` renders a page of 240 long links, streamed and prerendered, with the payload waiting from the start. It checks that every payload element follows a `>`, that the payload reads back whole, and that the document still ends with the held close.
  - Without the fix it fails: the streamed page's payload element lands after `…/a/piece/35">Page <!-- -->35<!--`, inside a comment.
  - With the fix, the file passes 8 of 8.
- `uf test packages/router`: 322 passed, 0 failed.
- `uf fmt --check` and `uf lint` on both files: clean.
- `uf check` on the two files reports 16 errors, all in lines this change does not touch: `ReadableStream`'s generics, the stream union in `preludeChunks`, a definition cycle in `bodyOf`, and `+` variance in the test's `handPayload`. It reports none in `advanced`, `Boundary` or the new case. CI's library check runs `uf lint`.
- A docs build with this change: the result follows in a comment.

Found on #1098, part of #947.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/1110"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792045818&installation_model_id=422833&pr_number=1110&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F1110&signature=fd57d9716a9ce8efa6bb2004db277bccf85200e37d0dd0c14e7b888a927ef412"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
A docs build from a branch carrying #1098 (the components reference, with every registry component rendered live), #1109 and this change, using a uf built from main:

- `tools/docs/build.sh`: succeeds, 56 pages.
- Payload elements inside markup: none on any of the 56 pages. Built from main without this change, every page had its data element inside markup.
- `tools/ci/links-resolve.sh`: passes. Without this change it reports `/gui<script type=` and `/refer<script type=`.
- `tools/ci/security-scan.sh`: `security-scan: ok`.

Issue #1140, opened 2026-09-15
react-compiler: use only the official compiler — replace uf_react_compiler's hand-written checks


ubugeeei:
The owner wants uf to use only the official React Compiler. Nothing in uf should reimplement a check the compiler makes.

## Where uf stands

**Official, and staying:**
- `uf build`, `uf dev` and `uf test` compile with `react_compiler` 0.1.0 (`crates/uf_transform/src/compiler.rs`, `compile_program`). The crate family is published from `facebook/react`'s `compiler/crates/`, by the workflow added in facebook/react#37465.
- `react/no-redundant-memo` (`crates/uf_transform/src/memo.rs`) asks that same compiler what it did.

**Hand-written, and to be replaced:** `crates/uf_react_compiler`
- About 4,700 lines, tests included. It is a validator uf wrote itself for "syntax mode", with its own scope model (`scope.rs`), token walker (`syntax.rs`), alias tracking (`analyze/`) and `useX` convention (`convention.rs`).
- It enforces five rules named after the compiler's validations: `react/hooks-rules`, `react/no-props-mutation`, `react/no-mutation-after-hook`, `react/no-ref-read-in-render` and `react/no-render-side-effects`.
- It never runs the official compiler. `official.rs` is a bridge to it that nothing calls.

Who uses the hand-written validator:
- `uf lint`: `crates/uf_lint/src/runner/react_compiler.rs` calls `uf_react_compiler::validate` for `react/hooks-rules` and `react/no-render-side-effects`.
- The `uf:react-compiler` plugin stage (`crates/uf_react_compiler/src/plugin.rs`), which refuses a module during a build. No production caller was found; confirm before removing it.
- The docs, which map ESLint's `rules-of-hooks` onto `react/hooks-rules` (`docs/app/guide/format/$page.mdx`).

## What to do

1. **Validate with the official compiler.**
   - `PluginOptions` has `output_mode: "lint"` (`CompilerOutputMode::Lint`) and `no_emit`. This is the approach `eslint-plugin-react-hooks` takes.
[...]

PR #1149, opened 2026-09-15
fix(transform): give the React Compiler each declaring identifier's binding, and check uf against its fixtures

ubugeeei:
Part of #1140

uf runs the official `react_compiler` crate, but the crate has no front end. It reads a Babel AST and a `ScopeInfo` built by whoever parsed the file. Upstream those come from `@babel/parser` and `@babel/traverse`; in uf they come from the Flow parser, `babel.rs` and `scope.rs`. This PR measures where uf's input differs from Babel's, against the compiler's own fixture corpus.

## Result

**1291 of 1809 fixtures pass (71.4%).** Every failure is attributed in `baseline.tsv`, and each note says whether its cause is proven or argued.

| bucket | fixtures | evidence |
|---|---:|---|
| conversion | 322 | **312 proven by experiment:** `babel.rs` sets neither `loc.filename` nor `loc.start.index`/`loc.end.index`. Babel's parser sets both, and the crate reads them for code frames and logger locations; adding them makes these pass. **5 proven by output or code:** Flow enums lowered to `$$ufEnum`; no `forwardRef` for a component's `ref` prop (React 19, `lower/components.rs`); the match fallthrough throw; the `import()` callee built without a location; a type annotation infer mode reads (`is_valid_props_annotation`). **5 argued:** 4 nested optional chains lose `?.` before the compiler (the crate's memo dependency drops it, which a printer cannot do); 1 capture seen only through a type annotation. |
| scope | 128 | **Proven by experiment:** `scope.rs` leaves each declaring identifier out of `ref_node_id_to_binding`. The Rust shim's `scope.ts` records it ("Map the binding identifier itself"), and recording it fixes all 128: 124 then pass, and 4 reach a later difference. |
| snapshot | 39 | **38 proven:** the snapshot also contains `babel-plugin-fbt`, `babel-plugin-fbt-runtime` or `babel-plugin-idx` output, because snap runs those plugins after the compiler in the same transform (`snap/src/compiler.ts`). **1 argued:** Prettier re-indented GraphQL inside a `graphql` template. |
| syntax | 28 | 8 the Flow parser rejects (`using`, `!`, instantiation expressions, `as` as an assignment target). 13 whose snapshot is a Babel parse error on input the Flow parser accepts. 3 TypeScript-only statements (`export =`, `import = require()`, `export as namespace`) that `print.rs` has no form for. 2 TypeScript `this` parameters, 1 TypeScript enum, 1 `@script`. |
| compiler | 1 | **Argued:** `timers.js`. uf hands the crate the JSX text as written, the crate's `trim_jsx_text` keeps a space, and the snapshot lacks it. The cause is either the crate or the snapshot's Prettier. |

No printer failures, and none left unattributed.

**On proving the compiler bucket:** facebook/react at the pin ships no Babel AST JSON. The `FIXTURE_JSON_DIR` data for `react_compiler_ast`'s tests is generated by `compiler/scripts/babel-ast-to-json.mjs`, which runs Babel. With no Babel allowed, anything attributed to the crate is argued unless the crate's code proves it.

## How it compares

`crates/uf_transform/tests/react_compiler_conformance/`:

- `pipeline.rs` runs uf's compile path: `lowered_ast`, `babel_from_lowered`, `scope::analyze`, `compile_program`, then `print`.
- `pragma.rs` gives each fixture the options its first line asks for. It ports `parseConfigPragmaForTests`, snap's `makePluginOptions` (including the shared-runtime type provider) and the Rust shim's `resolveOptions`.
- `normalize.rs` puts both sides into the same form before comparing:
  - **Code:** parsed by the official Flow parser, types erased, and compared as ESTree. Positions, comments, raw literal spellings, key quoting, trailing commas, empty statements and JSX text splitting are ignored, since a printer or Prettier chooses those.
  - **Errors:** compared as text, with the file name in `name.ts:3:4` lines replaced.
  - **Logs:** compared as JSON without `identifierName` and `fnLoc`, the two keys upstream's own `test-e2e.ts` drops.
  - Neither `print.rs` nor `babel.rs` touches the snapshot side, so a bug in either shows up as a difference.
- `expect.rs` reads `.expect.md` and ignores `### Eval output`.

## Pin, fetch and floor

- `tools/react-compiler/pin.txt` pins facebook/react `ff7445e6`. That is the commit `react_compiler` 0.1.0 was published from, per the crate's `.cargo_vcs_info.json`. A test checks the pin against `Cargo.lock` and, when the registry source is present, against that file.
- `tools/react-compiler/sync.sh` does a blob-less sparse checkout of the fixture directory only, into a folder named for the commit, and nothing is edited. `uf run react-compiler:conformance` syncs and runs the test.
- `baseline.tsv` records one line per fixture. The test fails when any fixture's result differs from it, so a fix has to record what it fixed and a regression can't merge silently. `UF_REACT_COMPILER_CONFORMANCE=update` rewrites the file. A new failure gets bucket `untriaged`, which the test refuses.
- CI: the Test, Upstream Flow Typecheck and npm publish jobs run the fixture sync before the suite. `tools/ci/workspace-suite-runtimes.sh` now refuses a suite job without that step, and two planted cases in its own test cover it.

## Cost

| step | time |
|---|---:|
| `time tools/react-compiler/sync.sh`, cold | 9.5 s |
| `cargo test -p uf_transform --test react_compiler_conformance -- --nocapture`, debug, 8 threads | 3.7 s (all 11 tests: 3.78 s) |

The CI figure will follow in a comment once the Test job has run.

## Babel

None, anywhere: no `@babel/*` dependency and no Babel run, in `packages/*`, crates, `tools/` or CI.

## Next, stacked on this

- **`scope.rs` declaring identifiers:** measured 1415 pass (+124, none lost).
- **After #1147 and the categories PR**, through one `compile_with_options`:
  - `loc.filename` and index: measured 1709. This also moves 19 error snapshots to `compiler`, proven by `react_compiler_diagnostics` `code_frame.rs`, which prints one filename for every detail.
  - Applying the compiler's renames: measured 1711.

## Not verified here

- `cargo clippy` was not run locally, to spare disk; CI's Clippy job will check it.
- CI wall time.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/1149"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792069432&installation_model_id=422833&pr_number=1149&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F1149&signature=f9ee03bc46faa5316476d7e88f3e560f041924764b5bfa59f9fc1c5253fb8162"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
#1150 was merged into this branch at 13:20Z, as 29e225af, so this PR now carries two changes:
- **The conformance run** over the React Compiler's own 1,809 fixtures.
- **The `scope.rs` fix.** It records each declaring identifier as a reference to its binding, as upstream's `scope.ts` does. This takes the run from 1,291 passing fixtures to 1,416 (+125, none lost).

The title now starts with `fix(transform)`, so the squash commit and the changelog name the fix rather than only the test.

PR #1172, opened 2026-09-16
fix(transform): give the compiler the positions and the name Babel gives it

ubugeeei:
The second of the two conformance fixes, through the entry #1166 added.

## What the compiler was missing

Babel's `loc` points carry an `index` beside the line and the column, and
its parser stamps `sourceFilename` onto the module's own `loc`. The React
Compiler reads both:

* the offsets are the range a suggested fix covers — `validate_exhaustive_dependencies.rs` attaches its inferred-dependency suggestion only when it has one to offer, and it has none without them;
* the name heads the code frame printed under a diagnostic — `program.rs` reads it once from `program.base.loc.filename` and carries it through `set_source_filename`.

uf set neither, so its errors arrived without the frames the snapshots
record and its logged locations without their offsets.

## Both follow the parser that read the module

This is the part that took the measuring. Whether either field exists is a
fact about the *parser*, not about uf: `@babel/parser` sets them and
`hermes-parser` does not. Three independent checks against the fixtures —
the oracle, since there is no Babel in this repository — agree:

| check | result |
|---|---|
| snapshots heading a diagnostic with `<stem>.ts:line:column` | **288** |
| snapshots heading one with a `.js` name | **0** |
| of the 16 snapshots carrying an inferred-dependency suggestion, Flow fixtures | **0** |

`pragma::filename` already encodes the same split: a TypeScript fixture is
compiled as `/<stem>.ts` and a Flow one as `/<stem>`.

So `babel.rs` writes `index` when the module's first line does not declare
Flow — `parseLanguage`'s own rule — and `compile_with_options` takes the
source file name as its own argument, which the crate deliberately keeps
apart from `options.filename`: *"the filename stored on AST node
`loc.filename` fields, which may differ from `filename`"*. uf's front end is
the Flow parser, so `uf build`, `uf lint` and the redundant-memo question
pass neither, and are unchanged.

## Measured

All 1,809 official fixtures at facebook/react@`ff7445e6`:

| | fixtures passing |
|---|---|
| main | 1,420 |
| this PR | **1,731 (95.7%)** |

**311 gains, 0 losses, 0 untriaged.** Counted from `baseline.tsv` against
`HEAD`, not from the run's changed-fixture list, which stops at sixty. The
run takes **4.3 s on 8 threads**; 111 `uf_transform` unit tests and 11
conformance tests pass.

Two unit tests hold the contract rather than the count: a non-Flow module's
`loc` points carry `index` equal to the node's own `start`/`end`, and a
Flow module's carry none.

### The 19 that nearly went into the baseline

Writing either field for *every* module instead looked better by one
measure — 1,712 — and was wrong. It regressed **19 Flow fixtures** that had
been passing: the compiler headed their diagnostics with a name and offered
fixes their snapshots do not have. They surfaced as `untriaged`, which is
exactly what would have been recorded as the new normal had the baseline
been updated on the strength of the headline number. They are gains in this
PR, not a bucket.

## Allocation budget

The budget fixture `packages/router/internal/runtime.js` begins `// @flow`,
so under this rule uf writes **no `index`** for it — the per-node cost the
ceilings guard is not paid on the module they measure. Both ceilings pass.

With this PR, `alloc_report` over that fixture: **795,777 allocations/run,
146.64 MiB/run, peak 19.38 MiB**, against 74,154 source bytes. The
unconditioned variant measured 808,871 and 148.95 MiB, so the gating costs
less than writing `index` everywhere, and both ceilings already held under
that heavier version. A before/after pair on the same run shape follows as
a comment; the figure in the budget's own comment (709,167) is not a
comparison — it was taken on a 68 KiB fixture.

Part of #1140

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/1172"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792131949&installation_model_id=422833&pr_number=1172&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F1172&signature=7c766373c569f19be69471b9481a09503f6694c559f8cef765cd153142303896"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


ubugeeei:
Allocation budget, before and after, on the same fixture with the same command (`cargo run --example alloc_report -p uf_lint -- packages/router/internal/runtime.js`, 74,154 source bytes both runs):

|  | allocations / run | bytes / run | peak |
|---|---|---|---|
| main `5123d50f` | 795,797 | 146.64 MiB | 19.38 MiB |
| this PR `54bb4481` | 795,777 | 146.64 MiB | 19.38 MiB |

Twenty allocations apart — run-to-run noise, and marginally lower after — with identical allocated volume and peak. **Neither ceiling moves and neither needs raising.**

That follows from the rule rather than from luck: `packages/router/internal/runtime.js` begins `// @flow`, so uf writes no `loc.*.index` for it. The per-node cost the ceilings guard is not paid on the module they measure. Both ceilings also passed under the earlier variant that wrote `index` for every module (808,871 allocations, 148.95 MiB), which is the heavier case.

The 709,167 in the budget's own comment is not a comparison: it was measured on a 68 KiB fixture, and this one is 74,154 bytes.

Issue #1177, opened 2026-09-16
release: `latest` is four different versions across the scope, and the packages pin each other exactly


ubugeeei:
Every `@uniflowed/*` package carries `0.0.0-alpha.39` on the `alpha` dist-tag. `latest` is **four different versions**, measured across every name in `tools/release/published-packages.txt` today:

| `latest` | packages |
| --- | --- |
| `0.0.0-alpha.11` (9) | core, host, router, server, stylex, test, ui, validator, vite |
| `0.0.0-alpha.12` (8) | config, fetch, graphql, hooks, query, react, react-testing, relay |
| `0.0.0-alpha.18` (2) | cell, state |
| `0.0.0-alpha.35` (1) | temporal |

## Why this is worse than "`latest` is behind"

Every `@uniflowed/*` pins its siblings exactly. So `npm i @uniflowed/react` with no tag does not install an old-but-working toolchain — it installs `react@0.0.0-alpha.12` asking for siblings at alpha.12 beside a `temporal@0.0.0-alpha.35` and a `core@0.0.0-alpha.11`. The set cannot agree with itself.

## What is NOT affected

`uf new` is fine. #408 fixed the scaffolder to write the running binary’s exact version (`"{uf}"`) rather than `"latest"`, and `crates/uf_project/src/tests.rs:198` pins it — the test fails if any manifest dependency is a dist-tag. A scaffolded project gets a coherent alpha.39. **This only bites someone who hand-writes `npm i @uniflowed/...` without a tag.**

## Why it drifted, and why it will drift again

`latest` sits wherever each package last published *under that tag*, so a release that skipped a package left that package behind. The skips are documented: #560 (`@uniflowed/temporal` was not publishable) and #1019 (`ui`, `state`, `cell` were not on npm). That explains the exact shape above — temporal at alpha.35, cell and state at alpha.18, the rest at alpha.11/12.

This is structural: unless `promote-latest` moves **every** package together, the next partial publish re-splits the tag.

## The decision, which is yours

`promote-latest` is a manual step needing your 2FA, so I cannot do either of these:

1. **Promote all twenty to `0.0.0-alpha.39`.** `latest` becomes coherent and matches `alpha`. The cost is that `latest` then points at a prerelease.
2. **Leave `latest` where it is and say so loudly** — have the README, the docs and the installer name `@alpha` explicitly, so nobody reaches for an untagged install.

[...]

Issue #1174, opened 2026-09-16
react-testing: IMPLICIT_ROLES is a second, hand-maintained copy of the HTML-AAM data the ARIA table now generates


ubugeeei:
Found while fixing the `<output>` regression on #1159.

`packages/react-testing/internal/queries.js` carries a hand-maintained `IMPLICIT_ROLES` map, which `getByRole` uses to decide what role an element has without an explicit `role` attribute. That is the same HTML-AAM data that `tools/aria/gen-table.cjs` now generates into `crates/uf_lint/src/runner/tree/aria/table.rs` from **aria-query 5.3.2** — so uf has two encodings of one specification, one generated and one typed by hand.

## Why it matters

The hand-maintained copy was already wrong. `output`'s implicit role is `status` unconditionally, and `IMPLICIT_ROLES` jumped straight from `option` to `progress` — so `getByRole("status")` could not find a bare `<output>`, and two story tests failed the moment a lint rule correctly told us to drop a redundant `role="status"`. One gap was found because something happened to trip over it. **Nobody has audited the rest**, and there is no reason to think it was the only one.

The two copies also drift in different directions: the generated table tracks whatever aria-query ships, while the hand-written map only changes when somebody notices a bug.

## Done when

- `IMPLICIT_ROLES` is reconciled against the generated table — every disagreement either fixed or recorded with the reason the testing library deliberately differs.
- Ideally there is one source: either the map is generated from the same aria-query data, or `getByRole` reads the generated table. If the two genuinely need different subsets, say so in a comment where the map is defined.
- A test pins the agreement, so a future aria-query bump cannot silently reopen the gap.

Worth knowing for whoever takes it: the story suite cannot exercise `@uniflowed/react-testing` from an agent worktree — it resolves out to the main checkout or a bun-cached release — so verify through the react-testing suite's own tests and treat CI's `Library` job as the authority.

ubugeeei:
One obstacle this issue names is already solved, so whoever picks it up need not re-derive it.

> the story suite cannot exercise `@uniflowed/react-testing` from an agent worktree — it resolves out to the main checkout or a bun-cached release

Agent worktrees live inside the main checkout, so `@uniflowed/*` climbs to the main checkout's `node_modules` and you end up testing main's code rather than your edits. Link the worktree's own packages in first:

```sh
mkdir -p <worktree>/node_modules/@uniflowed
ln -s <worktree>/packages/* <worktree>/node_modules/@uniflowed/
```

[...]

PR #1180, opened 2026-09-17
docs: tell package installs to use alpha

### LINKED ISSUE #1010: test: `uft.useFakeTimers` and `uft.useRealTimers` read as hooks, so `uf check` reports every test that controls the clock
## What happens

`packages/test/internal/timers.js` declares the fake clock as `export function useFakeTimers()` and `export function useRealTimers()`, and `uft` hands both on under those names. With component syntax on, Flow's React rules treat a function named `use` followed by a capital letter as a hook, so every call from a test's `it` callback — which is not a component or a hook — is reported:

```text
error[react-rule-hook]: Cannot call hook because React hooks can only be called within components or hooks.
```

On `main` at 45e291e9, `uf check packages/ui packages/react-testing --json` reports 37 of these, every one a test controlling the clock. No component or hook is involved in any of them.

## What does not fix it

Renaming at the call site does not help. `const { useFakeTimers: installFakeClock } = uft;` followed by `installFakeClock()` is reported exactly as before, so the report follows the declaration's type rather than the name the test calls it by. (Found while writing `packages/ui/interactions.test.js`, which has the same ten reports for the same reason.)

## What would

The declarations are the thing to change. `internal/isolation.js` already imports `useRealTimers as restoreRealClock` for readability; declaring the implementations under verb-first names (`installFakeClock`, `restoreRealClock`) and deciding separately what `uft` exposes — keeping the Vitest-compatible names as properties is what a migrating suite needs — would take the hook-shaped declaration out of the picture.

## How it will be tested

A type test in the style of `tests/type-tests/`, run through `everyMisuseIsReported`, that calls `uft.useFakeTimers()` and `uft.useRealTimers()` from a plain function and asserts `uf check` reports nothing for that fixture; and the `react-rule-hook` count for `packages/ui` dropping to zero.

### LINKED ISSUE #1017: react-testing: `screen` queries return `Element` and `userEvent` takes `HTMLElement`, so pressing a queried element is a type error
`@uniflowed/react-testing`'s two halves disagree about what an element is. The `screen` queries are typed to return `Element` (`packages/react-testing/internal/screen.js:82`), and every `userEvent` method takes an `HTMLElement` (`packages/react-testing/internal/events.js:629`, `click(element: HTMLElement, …)`). So the most ordinary line in a component test is a `uf check` error:

```js
await userEvent.click(screen.getByRole("button", { name: "Save" }));
```

```
error[incompatible-type]: Cannot call userEvent.click with screen.getByRole(...) bound to element
because Element [1] is incompatible with HTMLElement [2].
```

On `main`, `uf check packages/ui/ui.test.js` reports 475 errors, and 274 of them are this one. Nothing runs `uf check` over test files, which is why they are invisible and why they have grown.

A test presses what a role query finds, and nearly everything a role query finds is an HTML element; Testing Library's own types answer `HTMLElement` from the `ByRole` queries for that reason. An SVG found by role is the exception, and it is the one that deserves the wider type.

## Done when

- [ ] Pressing an element a `screen` query returned type-checks without a cast — either the queries answer `HTMLElement`, or `userEvent` accepts what they answer.
- [ ] `uf check packages/ui/ui.test.js` no longer reports those 274.

Found while writing the registry component tests for #947, which refine each element through a local `html(element)` helper in the meantime.

### LINKED ISSUE #1046: ui: ToggleGroup in `multiple` mode puts `aria-orientation` on `role="group"`, which axe reports as aria-allowed-attr (critical)
`packages/ui/toggle-group.js` renders `type="multiple"` as `role="group"` and writes `aria-orientation` on it (around line 196). ARIA 1.2 does not allow `aria-orientation` on `group`: the attribute is used on `scrollbar`, `select`, `separator`, `slider`, `tablist` and `toolbar`, and inherits into `listbox`, `menu`, `menubar`, `radiogroup`, `tree` and `treegrid`. axe-core reports it as a critical violation:

```
aria-allowed-attr (critical) — Elements must only use supported ARIA attributes
<div aria-label="Formatting" aria-orientation="horizontal" role="group" …>
```

It shows up in the first `toHaveNoAxeViolations` over a multiple-mode group. `type="single"` is unaffected, because it renders `RadioGroup`, and `radiogroup` does support the attribute.

Two ways to fix it, and they are not equivalent:

- **Drop `aria-orientation` from the group.** Smallest change, and the arrow keys keep working, but a reader is no longer told which arrows move between the buttons.
- **Render `role="toolbar"`.** The APG's pattern for a row of toggle buttons with one tab stop and arrow-key movement is a toolbar, and `toolbar` supports `aria-orientation`. It changes what a screen reader announces ("toolbar" rather than "group"), so it is a decision rather than a patch.

## Done when

- [ ] A multiple-mode `ToggleGroup.Root` passes `toHaveNoAxeViolations` with every rule enabled, with a test in `packages/ui/ui.test.js`.

Found while adding `toggle-group` to the `uf ui add` registry (#947), whose test disables `aria-allowed-attr` for this group only, naming this issue, until it is fixed.

### LINKED ISSUE #1047: ui: a disabled Pagination.Previous or Pagination.Next is an `<a>` with no `href` and an `aria-label`, which axe reports as aria-prohibited-attr
`packages/ui/pagination.js` drops `href` from a disabled `Pagination.Previous` or `Pagination.Next` and keeps its `aria-label` and `aria-disabled`. An `<a>` with no `href` has no role, it is `generic`, and ARIA 1.2 prohibits naming a generic element. axe-core reports it as serious:

```
aria-prohibited-attr (serious) — Elements must only use permitted ARIA attributes
<a aria-disabled="true" aria-label="Previous page" …>
```

It shows up in the first `toHaveNoAxeViolations` over a pagination whose first or last page is current, which is every paginated list sooner or later. What a reader gets is also worse than the audit says: the name is prohibited, so some screen readers read nothing at all where "Previous page, dimmed" belonged.

Two shapes that keep the end announced and unavailable:

- **`role="link"` with `aria-disabled="true"`** on the `<a>`, which is what the `render` branch already does (`role: disabled ? undefined : "link"`, the other way round). A link role permits the name and the disabled state.
- **A `<span>` with the words as text** rather than an `aria-label`, so the name is content and there is no ARIA to prohibit.

## Done when

- [ ] A pagination on its first page, and one on its last, pass `toHaveNoAxeViolations` with every rule enabled, with a test in `packages/ui/ui.test.js`.

Found while adding `pagination` to the `uf ui add` registry (#947). Its example sits on a middle page, and its test disables `aria-prohibited-attr` only for the disabled end, naming this issue.

### LINKED ISSUE #1059: react-testing: a name from content includes `aria-hidden` descendants
`accessibleName` in `packages/react-testing/internal/queries.js` names an element whose role takes its name from its content with `textOf(element)` (lines 599 to 603), which is the element's whole `textContent`. The accessible name computation leaves out descendants that are hidden, so text inside an `aria-hidden="true"` or `hidden` child is part of no name; here it is part of the name.

Found while writing the styled menu for #947. A row that draws its shortcut beside its label, and hides that drawing from assistive technology in favour of `aria-keyshortcuts`:

```html
<button role="menuitem" aria-keyshortcuts="Control+N">New file<span aria-hidden="true">Ctrl N</span></button>
```

A screen reader announces "New file". `getByRole("menuitem", { name: "New file" })` finds nothing, because the query computes "New fileCtrl N". The same happens to any button or link with a decorative `aria-hidden` icon that carries text, such as a glyph font.

## Done when

- A name from content skips descendants that are `hidden` or `aria-hidden="true"`, with a test for each.
- The registry menu's test queries a row with a shortcut by its name, rather than by `aria-keyshortcuts` as it does until this is fixed.

### LINKED ISSUE #1060: ui: menus in a menubar have no name, because `MenubarTrigger` never registers as their trigger
`Menu.Body` names itself with `aria-labelledby` pointing at `${menu.base}-trigger`, but only while `menu.triggered` is true (`packages/ui/menu.js`, line 385). `triggered` is set by `useTriggerRegistration` in `packages/ui/internal/menu-tree.js`, which `MenuTrigger` and `MenuSubTrigger` call (`menu.js`, lines 221 and 716). `MenubarTrigger` gives itself the same `${menu.base}-trigger` id and points `aria-controls` at the body, but does not call it, so every menu in a menubar opens as a `role="menu"` with no name.

Found while writing the styled menubar for #947: after a click on the File trigger, `getByRole("menu", { name: "File" })` finds nothing, where the same query passes for `Menu`.

## Done when

- `MenubarTrigger` calls `useTriggerRegistration(menu)`, and a menubar menu is named by its trigger, with a test.
- The registry menubar test asserts that name, rather than following `aria-controls` as it does until this is fixed.

### LINKED ISSUE #1074: ui: the table sort announcement is visible text beside the table, and no prop reaches it
`TableRoot` renders its announcement beside the `<table>` as `<div aria-atomic="true" aria-live="polite" data-uf-table-status="" role="status">{message}</div>` (`packages/ui/table.js`, line 208). Nothing hides it visually and no prop reaches it, so after a sort the announcement shows as a line of text under the table.

A page with a stylesheet can target `[data-uf-table-status]`. A component styled with `@uniflowed/stylex`, which is how the registry table for #947 is built, cannot reach an element it does not render. Compare `ComboboxStatus`, which is a part the caller renders, so the caller can keep it out of sight.

## Done when

- The table announcement is out of sight by default (the usual visually hidden rules), or a part or prop lets a caller style it, with a test.
- The registry table stops noting that its sort announcement shows under the table.

### LINKED ISSUE #1081: ui: a narrow-screen Sidebar panel and a collapsed item tooltip take no class
On a narrow screen, `SidebarBody` renders `<SheetBody aria-label={label}>{nav}</SheetBody>` (`packages/ui/sidebar.js`, around line 233). A caller class reaches the `<nav>` inside, because `...rest` goes there, and it does not reach the sheet panel around it. `SheetBody` renders `DialogBody` with only `data-side`, so a sidebar styled with `@uniflowed/stylex`, which is how the registry sidebar for #947 is built, has a panel with no look on a narrow screen.

The same holds for a collapsed `SidebarItem`. It wraps itself in `TooltipRoot`, `TooltipTrigger` and `TooltipBody` (around line 285), and nothing reaches that `TooltipBody` either.

Both are the shape of #1074: an element the part renders, that the caller cannot pass anything to.

## Done when

- A prop or a part lets a caller style the narrow-screen panel and the collapsed-item tooltip, with a test for each.
- The registry sidebar stops noting that both are out of its reach.

### LINKED ISSUE #1084: ui: calendar weekday headings are empty to axe, which reports empty-table-header
Each weekday heading `CalendarMonth` renders is `<th aria-label="Sunday" scope="col"><span aria-hidden="true">Sun</span></th>` (`packages/ui/calendar.js`, around line 408). The only text in the heading is hidden from assistive technology, and `aria-label` is not heading text, so axe-core reports `empty-table-header` (minor) for every weekday: "Table header text should not be empty".

Found while auditing the registry calendar and date picker for #947, where both tests now disable that one rule and name this issue.

A heading can keep the short name on screen and the whole name for a reader without hiding its text, for example `<th scope="col"><abbr title="Sunday">Sun</abbr></th>`, or the short name `aria-hidden` beside the whole name in visually hidden text.

## Done when

- The weekday headings pass `empty-table-header`, and a reader still hears the whole weekday name, with a test.
- The registry calendar and date-picker tests stop disabling `empty-table-header`.

### LINKED ISSUE #1144: build: a new project's first build warns INEFFECTIVE_DYNAMIC_IMPORT about the router's internals
`uf build` on a project from `uf new` prints two warnings. This was with `uf@0.0.0-alpha.37` and the packages installed from npm, found by the release e2e on 2026-09-15:

```
! [INEFFECTIVE_DYNAMIC_IMPORT] node_modules/@uniflowed/router/internal/error-view.js is dynamically imported by \0virtual:uf/client-references but also statically imported by node_modules/@uniflowed/router/internal/compose.js, node_modules/@uniflowed/router/internal/resolve.js, dynamic import will not move module into another chunk.

! [INEFFECTIVE_DYNAMIC_IMPORT] node_modules/@uniflowed/router/internal/boundaries.js is dynamically imported by \0virtual:uf/client-references but also statically imported by node_modules/@uniflowed/router/internal/compose.js, node_modules/@uniflowed/router/internal/runtime.js, dynamic import will not move module into another chunk.
```

**Why it matters**
- A user can't act on either warning. Both name uf's own virtual module and the router's internals.
- They appear on the first build of a fresh project, which is exactly where a toolchain should be quiet.

**Where it comes from**
- `virtual:uf/client-references` reaches every client reference through a dynamic `import()`.
- The router also imports these two modules statically, so the dynamic import can never split them into their own chunk, as the warning says.

**Done when**
- A fresh project builds without any warning that names uf's own modules.
- A test builds the `uf new` template and asserts that the build output carries no warnings.

One option is for `virtual:uf/client-references` not to emit a dynamic import for a module the graph already imports statically. Chunking wouldn't change, since that import has no effect today.

ubugeeei:
## Summary
- document that hand-installed `@uniflowed/*` packages should use the `@alpha` dist-tag
- carry the stacked accessibility, react-testing, timer, and Vite fixes now folded into this branch
- keep this as the bottom PR with auto-merge enabled

Closes #1177.
Closes #1174.
Closes #1084.
Closes #1047.
Closes #1046.
Closes #1060.
Closes #1059.
Closes #1074.
Closes #1081.
Closes #1010.
Closes #1144.
Closes #1017.

## Verification
- `./target/debug/uf fmt --check README.md tests/library/docs-tables.test.js`
- `./target/debug/uf lint README.md tests/library/docs-tables.test.js`
- `./target/debug/uf test tests/library/docs-tables.test.js`
- folded follow-up fixes were verified in their original stacked PR runs; CI remains the merge authority for this combined branch


ubugeeei:
Independent corroboration for the approach here: I audited `README.md` and every file under `docs/` on `main` for install commands naming an `@uniflowed` package, and there are **no bare untagged installs to fix**. Exactly two lines are install-adjacent, and the only package named is `@uniflowed/ui@alpha`:

```
docs/app/reference/packages/$page.mdx:148  install one by hand: `npm install @uniflowed/ui@alpha`
docs/app/$page.js:74                       ... every <code>@uniflowed/*</code> release on npm is a prerelease under the ...
```

So this PR is a **guard against a future mistake** rather than a fix for a present one, which matches the synthetic positive case in the test. That is the right shape, and the test earns its keep for exactly the reason the orphan check does — the defect it catches is the one nobody has made yet.

I checked that the filter could actually match before trusting the empty result: it was verified against `npm install @uniflowed/ui@alpha`, a bare `npm install @uniflowed/ui`, and a flag-separated `pnpm add --save-dev @uniflowed/core`. An earlier pass of mine used `\b` anchors under `git grep -E`, where POSIX ERE leaves them undefined, and silently matched nothing — so I redid it.

One gap worth knowing, not necessarily worth fixing here: `documents()` walks Markdown, so `docs/app/$page.js:74` — the runtime banner that tells readers every release is a prerelease — is outside the new check. If that line ever grows an install example, nothing will catch it. Scoping the test to Markdown is a defensible line; I mention it only so the boundary is deliberate rather than accidental.

For the record on #1177's remaining half: the incoherent `latest` itself is untouched by this — nine packages at `alpha.11`, eight at `alpha.12`, two at `alpha.18`, one at `alpha.35` — and only `promote-latest` moves it. This PR removes the documentation exposure; the improvising user who types `npm i @uniflowed/react` from memory is still the open case.

ubugeeei:
Heads-up rather than a nudge: **two PRs are now stacked behind this one**, so it is the gate for all three.

```
#1180  base main                         <- this, BLOCKED on 1 unresolved thread
  └─ #1182  base codex/issue-1177-alpha-docs        (Closes #1174)
       └─ #1184  base codex/issue-1174-implicit-roles (Closes #1084)
```

Both children are correctly **not** armed — auto-merge on a non-main base is the one thing that must never be on — so nothing moves until this merges and they re-target `main` in turn.

The only thing holding this PR is a single CodeRabbit thread (Minor, about the install-example regex in the new docs test); every check that has reported is green. I have not touched it — it is yours, and the finding is a judgement call about the regex rather than something I should decide for you.

Worth knowing either way: a review thread blocks the merge under `required_conversation_resolution`, so a PR can sit green and armed indefinitely on one unresolved comment. That happened to #1178 overnight for twelve hours before I spotted it. Whatever you decide on the thread — fix it or reply declining it with the reasoning — resolving it is what releases the stack.

ubugeeei:
I verified the blocking finding so you can decide it quickly. **It is real, but latent — nothing in the repo has the shape today**, so this is hardening rather than a live hole.

**The defect is real.** `untaggedInstalls` works per physical line — `INSTALL_COMMAND.test(line)`, then `line.matchAll(UNTAGGED_UNIFLOWED_PACKAGE)` on the same line. A shell continuation therefore splits into two halves that each pass:

```
npm install \
  @uniflowed/ui
```

Line 1 has the verb and no package; line 2 has the package and no verb. Neither is reported.

**But the shape does not occur.** Searching `main` for an install command ending in a backslash across `README.md` and `docs/**` returns nothing, and exactly one file contains any trailing-backslash line at all — `docs/app/guide/ci/$page.mdx`, a CI shell snippet, not an install. I checked the probe could actually match a continued install first, against a fixture containing one, so the empty result means something rather than being a pattern that could never fire.

So: **merging as-is leaves no known gap**, and the fix is about the example someone writes next year.

**One problem if you do take it.** The proposed regression test expects the package line wrapped in backticks:

```
{ line: 2, text: "`@uniflowed/ui`" }
```

That is asserting the diagnostic text for a Markdown-formatted continuation, which is a stranger shape than the plain `npm install \` + `@uniflowed/ui` the finding describes. If you implement continuation joining, I would pin the plain form and let the diagnostic report the *logical* command rather than whichever physical line the package landed on — otherwise the test encodes a formatting accident.

My read: this is a legitimate "fix it when convenient", not a merge blocker — but it is your call, and either resolving it or replying to decline it releases #1182, #1184 and #1185 behind you. I have not touched the PR.

For completeness, one thing I could not check from `main`: the scanner only exists on this branch, so I read it from the diff rather than the tree.

ubugeeei:
Verified the new blocking finding so you can decide it in one pass. **It is live and it is right** — unlike the last one, I would fix this before merging.

**Still present on `f48cf3ee`**, `packages/react-testing/internal/queries.js`:

```js
if (tag === "select") {
  return element.getAttribute("multiple") != null || element.getAttribute("size") != null
    ? "listbox"
    : "combobox";
}
```

`size` is tested for *presence*, so `<select size="1">` classifies as `listbox`. HTML-AAM maps a `select` to `combobox` unless `multiple` is present or the effective `size` is greater than 1 — so `getByRole("combobox")` misses the commonest select there is, and `getByRole("listbox")` finds one that is not. That is a real wrong answer from a role query, not a latent edge.

**One angle I checked and am ruling out, so nobody re-opens it:** this is *not* a second hand-maintained copy of generated data. `"select"` appears 13 times in `crates/uf_lint/src/runner/tree/aria/table.rs`, and every one is the ARIA **role** `select` inside an `ancestors` list (of `listbox`, `menu`, `menubar`) — not the HTML `<select>` element. The generated table does not encode the `size`/`multiple` condition, so there is nothing here for `queries.js` to read from and no #1174-style duplication to fix. (Same role-versus-element trap `marquee` sets in that file.)

**One caveat on the proposed patch**, worth a glance before taking it as written:

```js
const size = Number(element.getAttribute("size"));
... Number.isInteger(size) && size > 1
```

It silently decides what *invalid* values mean. `size=""` and `size="0"` fall to combobox, which is right. But `size="2.5"` is not an integer so it falls to combobox, where HTML clamps a malformed value rather than ignoring it, and `size="+3"` parses to `3` and becomes listbox, where HTML would reject the leading sign. Small next to the main defect, and arguably fine for a test-library heuristic — but it is a decision the patch makes rather than one the spec makes, so it deserves to be deliberate. If you want it strict, parse with the HTML integer rules; if you want it forgiving, a comment saying so would stop the next reader filing it as a bug.

Nothing else blocks this PR — every check that has reported is green, and #1195 and #1196 are stacked behind it.

Issue #1314, opened 2026-09-21
release: bind remaining implemented packages to trusted npm publishing


ubugeeei:
The scope page still needs one tracker for implemented packages that remain outside tools/release/published-packages.txt.\n\nThis is not the React Native target (#501): @uniflowed/react-native-testing is now in the published closure. Track the remaining package-publication work here:\n\n- bind the packages still listed in tools/release/pending-packages.txt to the publish workflow\n- keep tools/release/published-packages.txt closed under @uniflowed/* dependencies\n- rerun release preflight before moving each name\n\nThe gap row in docs/app/guide/scope/$page.mdx should point here until the pending list is empty or each remaining package has a deliberate non-publish decision.

Issue #1313, opened 2026-09-21
ui: number field parity


ubugeeei:
Split from #948.\n\nTrack a locale-aware NumberField primitive with the React Aria parity behaviours uf does not ship yet:\n\n- locale parsing and formatting\n- keyboard and pointer stepping\n- min/max/step validation\n- Field integration for labels, descriptions and errors\n\nThe parity table in docs/app/reference/ui/$page.mdx should point here until this row ships.

Issue #1312, opened 2026-09-21
ui: bidirectional and locale-aware collection behavior


ubugeeei:
Split from #948.\n\nTrack the cross-cutting React Aria parity work that sits below the visible primitives:\n\n- I18nProvider-style locale and direction propagation\n- RTL keyboard direction in roving-focus and collection components\n- collator-backed typeahead and sorting hooks where a component exposes them\n- tests that run the behaviours in at least one RTL locale\n\nThe parity table in docs/app/reference/ui/$page.mdx should point here until these rows ship.

Issue #1311, opened 2026-09-21
ui: range calendars and segmented date and time fields


ubugeeei:
Split from #948.\n\nCalendar and DatePicker cover the single-date path. Track the React Aria date/time rows that remain:\n\n- RangeCalendar\n- DateRangePicker\n- DateField\n- TimeField\n- segmented keyboard editing, announcement and validation semantics for each\n\nThe parity table in docs/app/reference/ui/$page.mdx should point here until these rows ship.

PR #1315, opened 2026-09-21
feat: add Windows release and editor publishing coverage

### LINKED ISSUE #267: router: no parallel routes, no intercepting routes, no template — and @slot and (.)segment currently serve as literal URLs
uf's route tree has exactly one page per URL and no way to render a route somewhere other than where its path says. Next's two conventions for that — parallel routes and intercepting routes — have no equivalent, and neither does `template.js`.

## What Next.js does

**Parallel routes.** A folder named `@team` is a *slot*: the segment's layout receives it as a prop alongside `children`, and several slots render at once in one view, each navigable independently and each with its own `loading` and `error`. `default.js` is what a slot renders when the current URL says nothing about it — which is what makes a hard navigation to a URL that only one slot knows about still produce a complete page. A slot folder never appears in the URL.

**Intercepting routes.** `(.)photo` beside a route matches `/photo` *from within this segment* and renders it here instead, leaving the URL as `/photo`. `(..)`, `(..)(..)` and `(...)` climb the tree. Together with a parallel route, this is the photo-in-a-modal pattern: clicking a thumbnail opens the photo over the feed, the URL is shareable, and loading that URL directly renders the full page instead.

**`template.js`.** A layout that remounts on navigation instead of persisting — for enter animations and per-navigation state, where `layout.js`'s persistence is the wrong default.

## What uf does today

Nothing of the three, and the grammar has no room reserved for them.

`packages/vite/internal/routes.js` `walk()` treats every directory as one path segment, with three cases at `routeFromSegments` (`routes.js:157`): `(group)` disappears, `[name]` is a parameter, `[...name]` is a catch-all, everything else is a literal. So:

- `@team` becomes the literal URL segment `/@team`;
- `(.)photo` is *not* a group — the test is `startsWith("(") && endsWith(")")` and `(.)photo` fails it — so it becomes the literal segment `/(.)photo`.

Both would be silently served as URLs. `crates/uf_router/src/lib.rs:200` has the same three cases and the same outcome, so `router.js` would generate `RoutePath` including `"/@team"`, and `route("/@team", ...)` would type-check.

`ReservedRole` (`crates/uf_router/src/reserved.rs`) has six roles and none of them is `default` or `template`. `RouteRecord` (`packages/router/internal/runtime.js:128`) is one `page` and an array of `layouts`, and `RouteView` (`runtime.js:709`) composes exactly that: one page, wrapped in its layouts innermost-last. There is no second child to give a layout and no place to put one.

## Why it matters here

The modal-with-a-real-URL is the pattern people name when they say an App Router application cannot be built on a simpler router, and it is one of the few Next.js features that genuinely changes what an application can be rather than how fast it is. The redundancy guide's framework row says "Target Next.js-class completeness", and this is the part of routing that is furthest from complete.

It is also the piece most worth designing rather than copying, which is why it is one issue and not four. Next's spelling is three kinds of parenthesis and an `@`, and the reason there is a `default.js` at all is that slots and URLs disagree about what is being rendered. uf has a stricter starting point: `crates/uf_router/src/reserved.rs` is a single grammar shared by the scaffold, the router and the linter, `router.js` generates a closed `RoutePath` union, and Modern Flow can state "this layout renders these named slots" with `renders` in a way a `default.js` convention exists to work around. A design that arrives at fewer files than Next has, and types the slots, would be a better answer rather than a copy.

The risk of leaving it is the cheap half: `@team` and `(.)photo` are currently *valid route directories that serve nonsense URLs*, so a project migrating from Next gets working-looking output rather than an error.

## The pieces

1. **Refuse what is not supported.** A directory named `@x` or `(.)x` inside the router root should be a `uf lint` error and a build failure, today, before any of the rest is designed. This is separable and small.
2. **Slots.** A route as a set of named children rather than one page: `RouteRecord`, `RouteView`, the layout props, and the generated Flow types for what a layout receives.
3. **`default`** — a seventh reserved role, and the resolution rule for a slot the URL does not address.
4. **Interception**, which needs a second matching pass that knows where a navigation came from, and a client-side navigation that can render one route while the URL says another.
5. **`template`**, which is small and independent: a reserved role and a `key` on the element in `RouteView`.

Doing (1) alone is worth a pull request on its own.

## Done

The grammar decision is written down in `crates/uf_router/src/reserved.rs`'s module docs, which is already where the reserved-name rules live and already explains why the two route implementations must agree. `crates/uf_router/tests/reserved_names.rs` covers the grammar; the rendering goes in `tests/library/`, and the interception case needs a client-navigation test rather than a render test.

### LINKED ISSUE #501: target: React Native, which the roadmap names twice and nothing consumes
`docs/roadmap.md` has "File-system router compiler for web and React Native targets" under P1 and "React DOM and React Native renderers" under P2, and `tools/upstream/repos.txt` pins React Native's codegen and Libraries. Nothing in the cargo graph or the packages reads any of it.

This is a target rather than a flag, and the work is in that order:

1. **What a target is.** `uf build --target native` has to mean something to the router compiler, the transform, the asset pipeline and the test runner, and each of those currently assumes a document.
2. **The renderer.** `@uniflowed/test` renders to a DOM; React Native needs the reconciler and a host config.
3. **The router.** A file-system router with no URL bar: the same route table, a different navigator.
4. **The transform.** Flow, the React Compiler and StyleX already run; Metro's transformer contract is the question.

**Done** is deliberately not one issue. This one is the entry point: it should be split as soon as the first of the four has an answer.

Part of #488.

### LINKED ISSUE #948: ui: React Aria parity — interactions, collections, and date and number fields
Measured against React Aria, `@uniflowed/ui`'s gaps are whole categories rather than single components:

- **Interactions.** `usePress` (pointer, keyboard and virtual clicks unified; a press released outside cancels), `useHover` that ignores emulated mouse events on touch, `useFocusRing`/`useFocusVisible`, `useLongPress`, `useMove`, and `useKeyboard` with propagation control.
- **Collections.** One collection model behind `ListBox`, `Menu`, `GridList`, `Table`, `Tree` and `TagGroup`: typeahead, multiple selection with shift ranges, disabled keys, sections, async loading, and virtualisation that holds at 10,000 rows.
- **Dates and times.** `Calendar`, `RangeCalendar`, `DateField`, `DatePicker` and `TimeField` on `@uniflowed/temporal`, with locale-aware segments.
- **Numbers.** `NumberField` parsing through `Intl.NumberFormat` — currency, percent, units — with step, minimum and maximum.
- **Drag and drop**, **Toast** with an announcing region, **ColorPicker**.
- **Internationalisation.** Right-to-left layout and keyboard direction, locale-aware collation when filtering.

## Done when

- [ ] `reference/ui` has a table of React Aria's hooks and components with uf's equivalent or "not yet", and each missing row has an issue.
- [ ] The interactions set ships first, tested over pointer, keyboard and virtual clicks.
- [ ] Every component shipped under this issue has keyboard-interaction tests and an axe assertion.

### LINKED ISSUE #968: platform: Windows — a native build, an install.ps1 that installs it, and a Windows CI job
There is no Windows build: `install.ps1` resolves and says so, WSL2 works, and a task written in shell syntax is refused without a POSIX shell. Vite+ treats Windows x64 as tier 1 (`irm https://vite.plus/ps1 | iex`). One Windows developer rules a toolchain out for the whole team.

## What to build

- `x86_64-pc-windows-msvc` release artifacts, and `aarch64` if the vendored Flow port builds there, with whatever linker and upstream patches that needs.
- `install.ps1` installing and verifying them the way `install.sh` does.
- Paths, process spawning, the `uf_env` and `uf_pm` stores (which lean on symlinks) and shell-less tasks working on Windows, or refused by name.
- A Windows CI job running `uf new`, `install`, `build`, `test`, `lint`, `fmt` and `check`.

## Done when

- [ ] The Windows job passes on every pull request that touches Rust.
- [ ] `guide/install` documents Windows with no WSL2 caveat.

### LINKED ISSUE #977: editors: publish the VS Code extension to the Marketplace and Open VSX, and make Zed work
`editors/vscode/README.md` says the extension is not published to any marketplace; `editors/zed` is a manifest and a README describing the missing Rust/WASM half; there is no JetBrains setup. Vite+ ships a VS Code extension pack, JetBrains support through the Oxc plugin, and Zed. An extension you have to build yourself is one nobody installs — and #941 and #949 add language-server features that only reach people who have it.

## What to build

- A release job packaging `editors/vscode` and publishing it to the VS Code Marketplace and Open VSX on each `uf@*` tag, versioned with uf. The publisher tokens are repository secrets the owner creates; a run without them skips by name.
- The Zed extension's Rust/WASM half, built and tested in CI.
- JetBrains: a documented LSP4IJ configuration at least.

## Done when

- [ ] A tagged release publishes the extension to both registries.
- [ ] Zed starts `uf lsp` for a uf project and shows diagnostics.
- [ ] The editors documentation covers installing from each.

ubugeeei:
## Summary
- add x86_64 Windows release packaging, installer verification, release smoke, and `uf toolchain` PowerShell installer execution
- add CI coverage for Windows project scaffolding and editor packaging, plus a tag-driven Editors workflow for VS Code Marketplace/Open VSX packaging and Zed WASM checks
- document the completed React Native, routing, installer, and editor release surfaces, with follow-up issues for remaining UI parity and package publication binding

Closes #948.
Closes #968.
Closes #977.
Closes #501.
Closes #267.

Follow-ups opened: #1309, #1310, #1311, #1312, #1313, #1314.

## Verification
- `tools/ci/gate-covers-every-job.sh`
- `sh -n tools/release/package-binaries.sh && sh -n tools/release/test-install.sh && sh -n infra/cloudflare/setup-assets/install.sh`
- `ruby -e 'require "yaml"; %w[.github/workflows/ci.yml .github/workflows/release.yml .github/workflows/editors.yml].each { |f| YAML.load_file(f); puts f }'`
- `cargo fmt --check`
- `cargo fmt --manifest-path editors/zed/Cargo.toml -- --check`
- `cargo check --manifest-path editors/zed/Cargo.toml --target wasm32-wasip1`
- `cargo build --manifest-path editors/zed/Cargo.toml --release --target wasm32-wasip1`
- `git diff --check`

Note: local `pwsh` is not installed on this machine, so PowerShell execution is covered by the added Windows GitHub Actions jobs.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/uf/pr/1315?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792569877&installation_model_id=422833&pr_number=1315&ref=codesmith_pr_footer&repository=ubugeeei-prod%2Fuf&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fuf%2Fpull%2F1315&signature=fd453cdd3116fc68fd9df8e76a3d76d30efdf64a181f4a4d5bbc6f4599478ee0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added native Windows x86-64 support, including PowerShell installation and packaged `uf`, `ufr`, and `ufx` executables.
  - Added editor integrations for Zed and JetBrains through LSP4IJ.
  - VS Code and Cursor extensions can now be distributed through Visual Studio Marketplace and Open VSX.
  - Added Windows release builds, installation verification, and checksum validation.

- **Documentation**
  - Updated installation, CI, editor, and integration guides for Windows and newly supported editors.
  - Added accessibility coverage for React Aria features.
  - Clarified routing and migration guidance.


ubugeeei:
Two separate failures on `2f14f24a`, one of which is measurably caused by this
PR. Posting because the PR has been red and unchanged for ~55 minutes.

## `Library` and `Library (Node 26)` — introduced by this PR

**Measured, not inferred:** the last six CI runs on `main` are all green,
including `d7c4d5b6` (#1316). These 8 failures are not pre-existing flake.

Both jobs fail on exactly the same 8 tests, both at 5m02s, 3350 passed — so it
is one cause, not two:

- `packages/hooks/hooks.test.js`
  - `useAsync, aborting and retrying > tries again … once it runs out of tries` — `expected 68 to be 3` (Node 26: `expected 71 to be 3`)
  - `Strict Mode … > does not start two intervals or two requests` — `timed out after 5000ms`
  - `useEventSource > reports the named event a message listener would have missed` — `getByText "open - - -": found nothing<div><output>connecting - - -</output></div>`
  - `useEventSource > stays out of the way while the browser is retrying by itself` — `expected 2 to be 1`
  - `useEventSource > opens a new connection when the browser has given up for good` — `expected 2 to be 1`
- `packages/ui/interactions.test.js`
  - `useLongPress > fires once the threshold passes …` — `expected ["longpressstart:touch"] to equal ["longpressstart:touch", "longpress:touch", "longpressend:touch"]`
  - `useLongPress > keeps the platform's own menu away …` — `expected true to be false`
- plus `packages/ui/ui.test.js` and `registry/ui/tabs.test.js` **exceeded the per-file budget of 300000000us**

**Inference, offered as a starting point rather than a conclusion:** every one
of these depends on time advancing — a retry count, an interval, an EventSource
reconnect, a long-press threshold — and two files then blow a 300-second
budget, which is what waiting on real time looks like. A retry that ran 68 times
instead of 3, and differed run to run (68 vs 71), reads like a loop that was
never clock-controlled. That points at fake timers not being installed rather
than at eight unrelated regressions.

What makes that worth checking first: the only runtime code this PR touches is
`packages/host/*` (`internal/node-hooks.js`, `internal/sync-hooks.js`,
`transform.js`) plus `crates/uf_cli/src/commands/toolchain.rs`. Everything else
is CI, editors, the installer and docs. So the host module-loading change is the
only plausible source, and that is also where `uft.useFakeTimers` would be
affected if module loading changed underneath it.

## `Windows` — the acorn failure is fixed, and the failure moved earlier

`b22d312e` worked: **zero** mentions of acorn in the new Windows log, where the
previous run died on `node_modules\acorn\dist\acorn.mjs`. But the job now fails
*sooner*. Comparing the two runs' own output:

| | previous (`b7e5f017`) | now (`2f14f24a`) |
| --- | --- | --- |
| `uf … ·` command banners | `uf new`, `uf install`, `uf test` | `uf new`, `uf install` |
| reached `uf test` | yes (`could not run 1 file`) | no |
| npm usage text printed | no | yes |

The tail is:

```
✓ dependencies installed in 22.95s
npm <command>

Usage:
…
##[error]Process completed with exit code 1.
```

So immediately after `uf install` reports success, **npm is invoked with no
subcommand** — that usage block plus exit 1 is what bare `npm` prints — and no
`uf fmt`, `lint`, `check`, `test` or `build` banner ever appears. uf logs its
own install as `command npm install --ignore-scripts --loglevel=http`, so this
is a *second*, separate npm invocation at or just after the end of `uf install`.

I could not find the code path that makes that second call, so I am not going to
guess at one — earlier on this PR I had a confident, plausible, code-level
hypothesis that turned out to be the wrong file entirely, and a diagnosis that
names the wrong place costs more than none. What is solid is the localisation:
`uf install` completes, something then runs `npm` with an empty subcommand, and
the step dies before any later command runs.

## Merge state

Independent of CI, there are **16 unresolved review threads**, and `main` has
`required_conversation_resolution` enabled — so the PR cannot merge while they
are open even once it is green. Auto-merge is armed, so it will go in by itself
once both are cleared.

